PR #3, opened 2019-11-16
Feature/unit testing

m4rtint:
I only Set up tests for Parking Indicator

so just need to look at
ParkingIndicatorTests
ParkingIndiactorBehavior
ParkingIndicator

NOTE:
I mixed up the class that needed the Behaviour name
If it's derived from MonoBehaviour -> attach the word Behaviour. 
ParkingIndicatorBehaviour : MonoBehaviour {} -> UI Stuff
ParkingIndicator {} -> Logic stuff


m4rtint:
[Unit Testing Setup](https://trello.com/c/v3flECRE/295-unit-testing-setup)

Assets/Script/Game/Car/ParkingIndicatorBehaviour.cs
```diff
@@ -0,0 +1,77 @@
+﻿//-----------------------------------------------------------------------
+// <copyright file="ParkingIndicator.cs" company="ANGRY GOOSE GAMES">
```
Gursaanj: Small thing, but the filename needs to be changed too

Assets/Script/Game/Car/ParkingIndicatorBehaviour.cs
```diff
@@ -0,0 +1,77 @@
...
+        set
+        {
+            m_target = value;
+        } 
+    }
+
+    private void Awake()
+    {
+        BaseWorldManager.OnFixedUpdate += ParkingIndicatorOnUpdate;
+        m_indicator = new ParkingIndicator(m_hideDistance);
```
wbeltran777: When we did the unit test meeting the behaviour script was the one not inheriting from monobehaviour, but in this file it's the other way around. Is there any reason why you switched it around ?

Assets/Script/Game/Car/ParkingIndicatorBehaviour.cs
```diff
@@ -0,0 +1,77 @@
...
+//-----------------------------------------------------------------------
+using System.Collections;
+using System.Collections.Generic;
+using UnityEngine;
+
+/// <summary>
+/// Simple parking indicator that points towards a target and disappears when player
+/// is within hideDistance to the target
+/// </summary>
+public class ParkingIndicatorBehaviour : MonoBehaviour
```
wbeltran777: Any reason why the naming convention got switched around? In the tutorial the "behaviour" script had all the logic and did not inherit from monobehaviour, but here the names are switched around.

PR #4, opened 2019-11-18
Feature/settings panel

Gursaanj:
Created Options Panel for the start menu. 

An issue is occuring with how the Audiomanager handles changes in volume (long story), but currently it is broken and changing the volume in the panel will not properly work, Will be addressed in a separate PR 

options panel


Gursaanj:
[Audio Settings Panel](https://trello.com/c/t7bUm23Z/287-audio-settings-panel)

Assets/Script/UI/MainMenu/QuitPanel.cs
```diff
@@ -19,7 +19,9 @@ public class QuitPanel : MonoBehaviour
     private Button m_no;
 
     /// <summary>Occurs when [on close panel].</summary>
-    public event Action OnClosePanel;
+    public event Action OnCloseQuitPanel;
+
+    public event Action PanelUpdate;
```
m4rtint: Having this delegate does not make sense.

Assets/Script/UI/MainMenu/OptionsPanel.cs
```diff
@@ -0,0 +1,214 @@
...
+
+    //Image the will highlight Text in the options panel
+    [SerializeField] private Image m_selectedOptionImage;
+
+    //Default values used in the option panel
+    private float m_defaultBGMVolume;
+    private float m_defaultSFXVolume;
+
+    public event Action OnCloseOptionsPanel;
+    public event Action PanelUpdate;
```
m4rtint: This delegate doesn't make sense

Assets/Script/UI/MainMenu/QuitPanel.cs
Gursaanj: I was trying to make a delegate where only when the Option panel (or Quit panel) is opened then we use the update method check if the back button is being pressed

Cause technically if you press the back button at the start menu you are still minimizing the panels even though its already minimized

What do you suggest?

Assets/Script/UI/MainMenu/OptionsPanel.cs
```diff
@@ -0,0 +1,214 @@
...
+        m_SFXSlider.onValueChanged.AddListener(ChangeSFXVolume);
+    }
+
+    private void Update()
+    {
+        PanelUpdate?.Invoke();
+    }
+}
+
+public class OptionPanelBehaviour
```
m4rtint: Swap the class names

Assets/Script/UI/MainMenu/OptionsPanel.cs
```diff
@@ -0,0 +1,214 @@
...
+{
+    #region Audio Management
+    /// <summary>
+    /// Updates the AudioManager's Background Volume
+    /// </summary>
+    /// <param name="volume"></param>
+    public void UpdateBackgroundMusicVolume(float volume)
+    {
+        AudioManager.Instance.MusicVolume = volume;
```
m4rtint: Store all the volume values inside playerprefs. So that when a player plays it again, the volume doesn't get reset

Assets/Script/UI/MainMenu/OptionsPanel.cs
```diff
@@ -0,0 +1,214 @@
...
+    }
+
+    private void Awake()
+    {
+        CheckDependancies();
+        transform.localScale = Vector3.zero;
+        m_selectedOptionImage.rectTransform.localScale = Vector3.zero;
+
+        // Get Event triggers of Option UI Elements
+        EventTrigger BGMSliderEventTrigger = m_BGMSlider.GetComponent<EventTrigger>();
```
m4rtint: It would be cleaner if you can somehow separate all this code into seperate methods.

Assets/Script/UI/MainMenu/QuitPanel.cs
Gursaanj: Would you rather in the update methods we have 

if (panel.isactive)
{
  Functionality();
}

instead of using the delegates?

m4rtint: go back to what I written. 

There is no issue with it trying to minimize if it's already minimized, since it checks if it's the correct size and just doesn nothing since it alraedy reached the given size. 

And this is not how you would use a delegate.

Assets/Script/UI/MainMenu/OptionPanelBehaviour.cs
```diff
@@ -0,0 +1,259 @@
...
+
+    /// <summary>
+    /// Gets the AudioManager's Background Volume
+    /// </summary>
+    /// <returns></returns>
+    public float InitialBackgroundMusicVolume()
+    {
+        if (PlayerPrefs.HasKey(m_backgroundVolumePref))
+        {
+            return PlayerPrefs.GetFloat(m_backgroundVolumePref);
```
m4rtint: No need for if statements here
https://docs.unity3d.com/ScriptReference/PlayerPrefs.GetFloat.html

just make the default value 1.0

m4rtint: Don't name this InitalBackgroundMusicVolume. Just "BackgroundMusicVolume" is fine.

"initial" makes people think it's the starting value when you first open the game.

Assets/Script/UI/MainMenu/OptionPanelBehaviour.cs
```diff
@@ -0,0 +1,259 @@
...
+
+    private void Update()
+    {
+        PanelUpdate?.Invoke();
+    }
+}
+
+public class OptionPanel
+{
+    private const string m_backgroundVolumePref = "BackgroundVolume";
```
m4rtint: Place these inside TagName - which btw, if we also make playerpref keys, should change tag name to something else.

Gursaanj: I was thinking of putting it in the MainText.cs file 

And just put all playerprefs in a region or something

Assets/Script/Data/MainText.cs
```diff
@@ -128,5 +128,14 @@ public static string AddOrdinal(int num)
         }
 
         #endregion
+
+        #region PlayerPrefKeys
+        
+        //All PlayerPrefKeys should be placed here
+
+        public const string BackgroundMusicVoumePref = "BackgroundVolume";
```
m4rtint: These should be in the file TagNames

m4rtint: The point of this file is to place UI Text here. Stuff that the users see

Assets/Script/UI/MainMenu/OptionPanelBehaviour.cs
```diff
@@ -0,0 +1,247 @@
...
+    {
+        m_BGMSlider.onValueChanged.AddListener(ChangeBGMVolume);
+        m_SFXSlider.onValueChanged.AddListener(ChangeSFXVolume);
+        m_BGMSlider.value = m_optionPanel.BackgroundMusicVolume();
+        m_SFXSlider.value = m_optionPanel.SoundEffectsVolume();
+    }
+
+    private void Update()
+    {
+        PanelUpdate?.Invoke();
```
m4rtint: Call OptionsPanelOnUpdate here
Delete PanelUpdate

*Note - Delegates are used for "DELEGATING" the function to SOMEONE ELSE.

Assets/Script/UI/MainMenu/QuitPanel.cs
```diff
@@ -46,35 +50,32 @@ private void SetButtonInteractable(bool isInteractable)
...
         SetButtonInteractable(false);
     }
 
     private void Update()
     {
-        if (AngryGooseGames.InputManager.IsCancelPressed())
-        {
-            ClosePanel();
-        }
+        PanelUpdate?.Invoke();
```
m4rtint: Call OptionsPanelOnUpdate here
Delete PanelUpdate

*Note - Delegates are used for "DELEGATING" the function to SOMEONE ELSE.

Assets/Script/Data/MainText.cs
Gursaanj: Fair enough, I will just change the TagNames files into something else, so it seems more appropriate to hold tag names, sorting layer names and PlayerPref Names

PR #5, opened 2019-11-20
Feature/spawn in players

m4rtint:
Because our code works so well, there really isn't much to change


m4rtint:
[Spawn the cars into the scene](https://trello.com/c/uIOUKmwz/304-spawn-the-cars-into-the-scene)

PR #7, opened 2019-11-23
Test/utilities

m4rtint:


mikeslgoh:
So why the decision to not keep it as a static class?

m4rtint:
@mikeslgoh you can't mock static classes

m4rtint:
Oh wait.. Hmm I see what you mean.
Yeah I don't havr an answer for that.

m4rtint:
I think. It's just better for testing.
Cuz with static class, there's no absolute clean way to null a static class.. Unless you create a custom class to clean out the state of the static class.

Assets/Tests/Utilities_Tests.cs
```diff
@@ -0,0 +1,258 @@
...
+    public class Utilities_Tests
+    {
+        private AngryGooseGames.Utilities utility;
+        [SetUp]
+        public void Setup()
+        {
+            utility = new AngryGooseGames.Utilities();
+        }
+
+        [TearDown]
```
Gursaanj: what Does [TearDown} do?

m4rtint: It's a method that is ran every single time a test has been completed. 
Lmk if u want more details on why or how to use it.

Gursaanj:
[Refactor Utility Class for testing](https://trello.com/c/jw2Q2YrJ/310-refactor-utility-class-for-testing)

PR #8, opened 2019-11-26
Balancing/car driving

m4rtint:


m4rtint:
Since I refactored inputmanager, i'll wait for that to be merged first before merging this one.

PR #10, opened 2019-11-28
Player Manager refactoring warnings

m4rtint:


m4rtint:
[Refactor PlayerManager](https://trello.com/c/PjdD5szT/317-refactor-playermanager)

Assets/Script/Managers/PlayerManager.cs
```diff
@@ -136,15 +137,9 @@ private void SetupPlayerCount()
...
-        try
+        if (m_players.Count < 5 && m_players.Count > 0)
         {
-            Assert.IsTrue(m_players.Count < 5 && m_players.Count > 0);
-        }
-        catch (AssertionException e)
-        {
-            Debug.LogError(e.Message);
-            Debug.LogError("Number of players needs to be between 1-4");
+            throw new Exception(Warnings.NumberOfPlayersWarning);
```
Gursaanj: Doesn't this mean that the Exception will be thrown if player count is between 1-4??

m4rtint: Yes, changing

Assets/Script/Managers/PlayerManager.cs
```diff
@@ -136,15 +137,9 @@ private void SetupPlayerCount()
 
     private void CheckNumberOfPlayers()
     {
-        Assert.raiseExceptions = true;
-        try
+        if (m_players.Count > 4 && m_players.Count <= 0)
```
Gursaanj: Shouldn't this be an OR statement, rather than an AND statement ?

PR #22, opened 2019-12-01
Powerup/boost

mikeslgoh:
Last feature to add from build day:
- Added Boost to object pooler and equip with key 5
- Made sure every powerup has an icon
- Renamed the project to PartyParkade
- Also made it that the image and text will be hidden when there is no powerup equipped


Assets/Script/ScriptableObject/Boost.asset
```diff
@@ -10,8 +10,8 @@ MonoBehaviour:
   m_Enabled: 1
   m_EditorHideFlags: 0
   m_Script: {fileID: 11500000, guid: 9f37146a2dab840debabee1ee0fb9738, type: 3}
-  m_Name: BouncyObjectPowerUp
+  m_Name: Boost
```
m4rtint: How come this asset is renamed?

m4rtint: Oh nvm

Assets/Script/UI/PlayerUserInterface.cs
```diff
@@ -70,6 +70,15 @@ public void SetPowerUpUI (PowerUpProperties properties)
         {
             powerName = properties.Name;
             powerUpSprite = properties.Icon;
+            if ((int)m_IconOfPowerUp.color.a == 0)
+            {
+                m_IconOfPowerUp.color = new Color(1,1,1,1);
```
m4rtint: Is it not possible to make the sprite/image null? 
instead of making it alpha?

mikeslgoh: It’s not really the image being empty I believe. It seems that the image default is a white background which is why I tried changing it to be transparent.

Assets/Script/UI/PlayerUserInterface.cs
```diff
@@ -66,14 +66,17 @@ public void SetPowerUpUI (PowerUpProperties properties)
     {
         string powerName = string.Empty;
         Sprite powerUpSprite = null;
+        Color spriteColor = Color.clear;
```
mikeslgoh: nice addition @m4rtint

PR #24, opened 2019-12-02
Stylecop and refactor object pooler

m4rtint:


Assets/Script/Game/Inventory.cs
```diff
@@ -45,46 +45,45 @@ private void OnDestroy()
...
         }
 
         if (Input.GetKeyUp(KeyCode.Alpha2))
         {
             Discard();
             powerUpTag = StoredStrings.GravityWell;
-            GameObject powerUp = ObjectPooler.Instance.SpawnFromPool(powerUpTag, SpawnPosition, Quaternion.identity);
-            Equip(powerUp.GetComponent<GravityWellBehaviour>());
+            PowerUp powerUp = PowerUpsObjectPooler.Instance.SpawnFromPool(POWERUPS.GravityWell);
+            Equip(powerUp);
```
Gursaanj: Put all of this into its own method if possible; alot of repeated code

Assets/Script/Game/Inventory.cs
```diff
@@ -45,46 +45,45 @@ private void OnDestroy()
     #region Debug
     private void AttachPowerUp()
     {
-        Vector3 SpawnPosition = transform.position + Vector3.back * 10;
         string powerUpTag;
         if (Input.GetKeyUp(KeyCode.Alpha1))
```
Gursaanj: Instead of if statements, would switch cases be better ?

Assets/Script/Game/Inventory.cs
Gursaanj: This was addressed in #25 my bad

Assets/Script/Game/Inventory.cs
mikeslgoh: Well for switch cases, you would need a variable to compare to where as for this, it's multiple different inputs? So might be hard to create a switch case for this

Gursaanj: yeah seems like it will be a lot of trouble, cause you can't make a switch case of KeyCodes since its looking for bools. Fair enough

PR #25, opened 2019-12-02
Refactor/debug player inventory

m4rtint:


Assets/DebugInventory.cs
```diff
@@ -0,0 +1,69 @@
+﻿using UnityEngine;
```
Gursaanj: Missing Header

m4rtint: I'll come back to it, i promise :P

PR #26, opened 2019-12-02
Bug fix/working volume control

Gursaanj:
Replaced AudioManager asset (Papae-AudioManager) to SoundManagerPro3. all changes were to accommodate for the new asset.

There are a lot of unnecessary files that came with the manager (i.e demo scenes, sounds effects) that I have kept in so that people can try out the demo if need be. We can always delete later

**Files that I changed:**

- AudioManager.cs
- OptionBehaviour.cs


Gursaanj:
[AudioManager Can't adjust volume](https://trello.com/c/VUiAiwG3/296-audiomanager-cant-adjust-volume)

Assets/Script/Managers/WorldManagers/BaseWorldManager.cs
```diff
@@ -90,7 +89,7 @@ private void PauseOrPlayGame()
         if (checkButtonPress || checkState)
```
Gursaanj: Im not sure if its because I made a mistake with resolving the merge conflict or not. 

But Im pretty sure this should be an  && instead of || ? 
Cause now the games is constantly just pausing and unpausing during the scene haha. Or am I missing something

m4rtint: Yes you're right, can you change it to be an &&

if used to be
if (checkButtonPress) {
    if (checkState){ 
    }
}

I just combined it

Assets/Script/Managers/AudioManager.cs
```diff
@@ -1,18 +1,18 @@
...
-using Papae.UnitySDK.Managers;
 using UnityEngine;
 using UnityEngine.SceneManagement;
 
+
 /// <summary>Class SoundManager.
 /// Implements the <see cref="UnityEngine.MonoBehaviour"/></summary>
 /// <seealso cref="UnityEngine.MonoBehaviour" />
-public class SoundManager : MonoBehaviour
+public class AudioManager : MonoBehaviour
```
m4rtint: Please Add Our namespace

Assets/Script/Managers/AudioManager.cs
```diff
@@ -37,7 +37,28 @@ private AudioClip RandomAudioClipFromArray(AudioClip[] soundEffectsArray)
     public void PlayOneShotSFX(SOUNDEFFECT soundEffect, float relativeVolume = 1.0f)
     {
         AudioClip sfx = RandomAudioClipFromArray(m_dictionaryOfSFX[soundEffect].AudioClipArray);
-        AudioManager.Instance.PlayOneShot(sfx, transform.position, relativeVolume);
+        float relativeSFXVolume = relativeVolume * PlayerPrefs.GetFloat(StoredStrings.SoundEffectsVolumePref); // get relative volume based on SFX volume set in options
```
m4rtint: Can you cache this ..aka store this in a local variable? 
PlayerPrefs.GetFloat(StoredStrings.SoundEffectsVolumePref);

I think, don't @ me. Whenever you call PlayerPrefs, it's the same thing as loading from hard disk.

mikeslgoh: Yeah especially if the value isn't going to change after START, might as well declare a local variable

Gursaanj: oh really?, alright Ill probably store it statically in the optionsBehaviour and reference that?

Gursaanj: > Yeah especially if the value isn't going to change after START, might as well declare a local variable

The thing is that people can change the SFX volume in the options menu at any time, so its not just after start, I think Ill just make it a variable that updates everytime the sfx slider changes

m4rtint: > oh really?, alright Ill probably store it statically in the optionsBehaviour and reference that?

You can store it in our audio manager. and just reference it there.

Gursaanj: NVM the soundmanager Asset has a method to get the SFX volume haha

m4rtint: Does it get stored in its internal library? if so , just use that instead of playerprefs

Assets/Script/Managers/AudioManager.cs
```diff
@@ -1,116 +1,118 @@
...
 
-    //Check to make sure that all Scenes in build are accounted for in dictionary of BGM
-    private void CheckForCorrectNumberOfScenes()
-    {
-        int numberOfScenesInBuildSettings = SceneManager.sceneCountInBuildSettings;
-        int numberOfScenesInBGMDictionary = m_dictionaryOfBGM.Count;
+        private AudioClip RandomAudioClipFromArray(AudioClip[] soundEffectsArray)
+        {
+            AudioClip sfx = soundEffectsArray[UnityEngine.Random.Range(0, soundEffectsArray.Length)];
+            return sfx;
```
m4rtint: Nitpick but you can just condense this to one line. 
return soundEffectsArray[UnityEngine.Random.Range(0, soundEffectsArray.Length)];
Don't need to change if u don't want to

Assets/Script/Managers/AudioManager.cs
Gursaanj: "private float _maxSFXVolume = 1f;" Nah it just defaults to 1 unless touched..

If Im reading this correctly

Assets/Script/Managers/AudioManager.cs
```diff
@@ -0,0 +1,118 @@
...
+        private void SetUpBackgroundMusicConnections()
+        {
+            foreach (SCENENAME scene in m_dictionaryOfBGM.Keys)
+            {
+                SoundConnection soundConnection = SoundManager.CreateSoundConnection(SceneInformation.NameOfMap(scene), m_dictionaryOfBGM[scene].AudioClipArray);
+                SoundManager.AddSoundConnection(soundConnection);
+            }
+        }
+
+        private void PlaySceneSoundConnection(Scene scene, LoadSceneMode mode)
```
wbeltran777: Why are we passing 'mode' here?

Assets/Script/Managers/AudioManager.cs
m4rtint: okay

Assets/Script/Managers/AudioManager.cs
Gursaanj: So unity has an event SceneManager.sceneLoaded which subscribes when a new level is loaded. Which we use to change the background music. 

The downside to using this is the event has the parameters of Scene and LoadSceneMode. We want scene but have no use for the other parameter at the moment (might find a use in the future)?

m4rtint:
The Audio Manager asset has a bunch of Demo Mp3 files. it takes up a bunch of space, i think u should be able to delete them.

PR #28, opened 2019-12-07
Balance/car ball

m4rtint:


m4rtint:
[Car Controller Ball Balance](https://trello.com/c/TwBJj8N2/330-car-controller-ball-balance)

m4rtint:
[Car Controller Ball Balance](https://trello.com/c/TwBJj8N2/330-car-controller-ball-balance)

Assets/Script/Game/Car/CarController.cs
```diff
@@ -105,117 +136,173 @@ private void OnDestroy()
...
-        Warnings.CheckObject(m_centerOfMass, WarningType.MissingChildReference, GetType());
         //Warnings.CheckObject(m_indicator, WarningType.MissingChildReference, GetType());
     }
 
-    /// <summary>
-    /// Check for inputs from InputManager.
-    /// Modify car acceleration accordingly.
-    /// </summary>
-    private void CarControllerOnUpdate()
+    private void Update()
```
wbeltran777: Any reason why we're not doing a global update here? Is it just for testing purposes in case we don't have world manager working?

m4rtint: Just straight up forgot.

m4rtint: okay Fixed

Assets/Script/Game/Car/CarAudioSystem.cs
```diff
@@ -20,7 +19,7 @@ private void Awake()
 
     private void Update()
     {
-        OnHornPressed();
+        //OnHornPressed();
```
mikeslgoh: Any reason why this is commented out?

Assets/Script/Game/Car/CarController.cs
```diff
@@ -79,143 +121,190 @@ public void Initialize(Vector3 position, PlayerInformation information)
...
+        transform.rotation = Quaternion.Slerp(transform.rotation, Quaternion.Euler(new Vector3(0, transform.eulerAngles.y + m_rotateTarget, 0)), Time.deltaTime * 2.0f);
     }
 
-    private void HandBrake(bool isBraking)
+    private void JumpIfNeeded()
     {
-        // Handbrake
-        if (isBraking)
+        // Jump
+        if (m_carProperties.JumpAbility && m_nearGround && Input.GetKeyDown(KeyCode.Space))
```
mikeslgoh: wait so the car can jump now? :O

m4rtint: If you allow it. but it's still disabled. and there for use if you enable it

Assets/Script/Game/Car/CarAudioSystem.cs
m4rtint: The sound prefab is still not on the scene yet.

Assets/Script/Game/Car/CarController.cs
mikeslgoh: Ok might be fun to play around with when we do a proper playtesting session

Assets/Script/Game/Car/CarController.cs
```diff
@@ -79,143 +121,190 @@ public void Initialize(Vector3 position, PlayerInformation information)
...
+        m_rotate = 0f;
 
-        foreach (WheelCollider wheel in m_wheels)
+        if (m_nearGround || m_carProperties.CanSteerInAir)
         {
-            wheel.brakeTorque = 0;
+            m_rotate = GetHorizontalMovement() * m_carProperties.Steering;
         }
+
+        transform.rotation = Quaternion.Slerp(transform.rotation, Quaternion.Euler(new Vector3(0, transform.eulerAngles.y + m_rotateTarget, 0)), Time.deltaTime * 2.0f);
```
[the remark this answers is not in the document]
m4rtint: Most of them don't matter and will just make it more confusing than needed. I only made the variables that matter available to the tester to change.

Assets/Script/Game/Car/CarController.cs
```diff
@@ -79,143 +121,190 @@ public void Initialize(Vector3 position, PlayerInformation information)
...
 
-    /// <summary>
-    /// Check for inputs from InputManager.
-    /// Modify car acceleration accordingly.
-    /// </summary>
     private void CarControllerOnUpdate()
     {
-        float horizontal = m_inputManager.HorizontalMovement();
-        bool isBraking = m_inputManager.IsHandBraking();
-        bool isDrifting = m_inputManager.IsDrifting();
```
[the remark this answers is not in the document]
m4rtint: No, the car driving is fine on itself. Unless there's a specific way of drifting is needed

Assets/Script/Game/Car/SmokeParticles.cs
```diff
@@ -0,0 +1,55 @@
...
+    private void ParticleEffects()
+    {
+        // Effects
+        if (!m_controller.CarPropery.MotorCycleTilt)
+        {
+            m_smoke.transform.localPosition = new Vector3(-m_controller.RotateTarget / 100, m_smoke.transform.localPosition.y, m_smoke.transform.localPosition.z);
+        }
+
+        bool isAcceleratingFastEnough = Vector3.Angle(m_controller.Sphere.velocity, m_controller.VehicleModel.forward) > 30.0f;
+        bool isMagnitudeEnough = m_controller.Sphere.velocity.magnitude > (m_controller.CarPropery.Acceleration / 200);
```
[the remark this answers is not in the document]
m4rtint: Same comment, these variables are more so for translating a magnitude to a value that makes more sense. The more variables, the more memory used. These are one of the variables that don't need to be changed or stored for any reason.

Assets/Script/Game/Car/CarController.cs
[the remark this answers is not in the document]
Gursaanj: How are you deciding which variables matter for the tester?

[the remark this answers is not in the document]
m4rtint: It's whatever was available in the original third party script. Just a simple single value that testers can change without breaking anything else. 
If there's anything that you can fine tune even more, then sure we can change that in the future. But for now, make it as easy as possible instead of it turning into the problem we had before with the wheel collider car controller with so many variables that you just can't tune it. 

The core problem is that the car didn't feel good, The values that are made public can be changed enough that it won't break anything.

PR #29, opened 2019-12-07
Created Scene with buildings and lighting

m4rtint:
Just a bunch of buildings and skybox added
Capture


m4rtint:
[Atomspheric Level Design](https://trello.com/c/YFtgGGxb/331-atomspheric-level-design)

PR #33, opened 2019-12-19
Black hole

DJSteins:
Fixes the black hole power up.


DJSteins:
[Refine blackhole powerup](https://trello.com/c/KhkLYkZw/337-refine-blackhole-powerup)

Assets/Script/PowerUp/GravityWellBehaviour.cs
```diff
@@ -71,23 +71,16 @@ private void OnTriggerEnter(Collider other)
...
-                {
-                    possiblePlayer = carController.PlayerInformation.PlayerNumber;
-                }
-                catch (NullReferenceException ex)
-                {
-                    print("NullReferenceException caught in GravityWellBehevaiour.OnTriggerEnter");
-                    return;
-                }
+                // appears to be the only way to get the car controller out of the collider
+                PlayerController playerController = other.gameObject.transform.parent.parent.gameObject.GetComponent<PlayerController>();
```
m4rtint: Might be better to use this
https://docs.unity3d.com/ScriptReference/GameObject.GetComponentInParent.html

DJSteins: Tried that and it didn't work :(

DJSteins: Actually I tried it with CarController... maybe it works with PlayerController though.

Assets/Script/PowerUp/GravityWellBehaviour.cs
```diff
@@ -71,23 +71,15 @@ private void OnTriggerEnter(Collider other)
...
-                {
-                    possiblePlayer = carController.PlayerInformation.PlayerNumber;
-                }
-                catch (NullReferenceException ex)
-                {
-                    print("NullReferenceException caught in GravityWellBehevaiour.OnTriggerEnter");
-                    return;
-                }
+                PlayerController playerController = other.GetComponentInParent<PlayerController>();
+                CarController carController = playerController.Car;
```
m4rtint: Why not just do GetComponentInParent<CarController> ?

mikeslgoh: @DJSteins do you have time to change this so that we're able to merge this?

DJSteins: I get the following error when I try that:
`NullReferenceException: Object reference not set to an instance of an object
GravityWellBehaviour.OnTriggerEnter (UnityEngine.Collider other) (at Assets/Script/PowerUp/GravityWellBehaviour.cs:75)`

DJSteins: Specifically trying `CarController carController = other.GetComponentInParent<CarController>();` does not work

PR #34, opened 2019-12-29
Parking Detection

DJSteins:
- Added two test parking spots to map
- Changed parking detection to use the bounding box of the car body
- Fixed `ScoreManager` because its initialization step was not being called in the `BaseWorldManager Start` function
- Added `ParkingIndicatorBehaviour` as a serialized field to CarController
- Changed the parking spots to have a box collider that is only as big as the orange part of the spot and extends high enough to cover the whole car size


DJSteins:
Not sure if we want to do more with this PR such as testing other parking detection mechanisms and/or parameters.

Assets/Script/Game/ParkingDetection.cs
```diff
@@ -42,24 +42,24 @@ public bool IsChosenToBeParkingSpot
...
-        foreach (Vector3 vertex in carVerticies)
+        // iterate over all the corners of the box collider to see if any are not within the parking spot
+        int[] signs = {1, -1};
+        foreach (int firstSign in signs)
         {
-            if (!parkingSpotBounds.Contains(vertex))
+            foreach (int secondSign in signs)
             {
-                return false;
+                Vector3 offset = new Vector3(firstSign * carEdges.x, 0.0f, secondSign * carEdges.z);
```
m4rtint: why is it iterating over
1, -1
-1, -1
1, 1
-1, 1

would it always going to be  +/- 1

Assets/Script/Game/ParkingTimer.cs
```diff
@@ -121,7 +121,11 @@ private void ResetVariables()
     private void OnParkingCompleted()
     {
         ScoreManager.CarScored(m_currentCarOnSpot);
-        PerigonGames.AudioManager.Instance.PlayOneShotSFX(SOUNDEFFECT.Parked);
+        // TODO: Resolve null reference
+        if (PerigonGames.AudioManager.Instance != null)
+        {
+            PerigonGames.AudioManager.Instance.PlayOneShotSFX(SOUNDEFFECT.Parked);
```
m4rtint: you could do 
PerigonGames.AudioManager.Instance?.PlayerOneShot

Assets/Script/Game/ParkingDetection.cs
DJSteins: In my opinion it felt less cumbersome to have the nested loop than what was previously been written.

Assets/Script/Managers/WorldManagers/BaseWorldManager.cs
```diff
@@ -42,15 +42,14 @@ private void Awake()
         {
             m_playerManager = GetComponent<PlayerManager>();
         }
-
         m_parkingManager.OnParkingChange += SetNewParkingSpot;
+
```
m4rtint: yeah sorry the scoremanager is not set up yet, since we haven't init the BaseWorldManager yet.
but it should be init after parking detection is ready.

Assets/Script/Game/ParkingTimer.cs
DJSteins: Shouldn't we leave as a TODO until we properly implement the audio manager?

Assets/Script/Game/ParkingDetection.cs
m4rtint: But is it always from -1...1

Assets/Script/Game/ParkingTimer.cs
m4rtint: A Prefab is supposed to be on the map. 
It's already implemented, it just need needs to be dragged and dropped onto the scene afaik 
@Gursaanj

m4rtint:
@DJSteins AFAIK, it's mostly ad hoc testing, and just basically recording and posting a video of what you've done in the PR

mikeslgoh:
> @DJSteins AFAIK, it's mostly ad hoc testing, and just basically recording and posting a video of what you've done in the PR

Yeah just add a video of the changes you made for this PR would be great

Assets/Script/Game/ParkingDetection.cs
DJSteins: Yeah but previously if you wanted to change what the offset was you would have had to change it in eight places (i.e. twice for each vertex). Now you just change it on one line so it is only two changes, therefore it is more maintainable code.

Assets/Script/Managers/WorldManagers/BaseWorldManager.cs
DJSteins: Should I put the `ScoreSetup` call back in `Start` then?

m4rtint: Yes plz

Assets/Script/Game/ParkingDetection.cs
m4rtint: Okay looking through logic again I understand the math.
I thought it was doing base +1, base - 1
Not base + (-1 *edge) 
Base +(1 *edge)

DJSteins:
@m4rtint video is on coding channel :)

mikeslgoh:
[Rework parking detection](https://trello.com/c/BBIvZjw1/338-rework-parking-detection)

PR #40, opened 2020-01-12
Black Hole Particle Effects and Parking Indicator

DJSteins:
- Adds a prefab that uses Unity particle effects for the black hole power-up
- Adds a reactive parking indicator


DJSteins:
[Black hole and parking indicator](https://trello.com/c/xurUYRCZ/345-black-hole-and-parking-indicator)

Assets/Epic Toon FX/Materials/Portals/portal_stripes_soft.mat
```diff
@@ -24,6 +24,6 @@ Material:
         m_Scale: {x: 1, y: 1}
         m_Offset: {x: 0, y: 0}
     m_Floats:
-    - _InvFade: 1
+    - _InvFade: 0.75
```
m4rtint: Just keep in mind that the materials are universal, so if you want to use the original one, you'll need to revert back to the original material properties

Assets/Epic Toon FX/Prefabs/Interactive/Powerups/PowerOrb/PowerOrbBlue.prefab
```diff
@@ -5124,7 +5124,7 @@ ParticleSystem:
         m_RotationOrder: 4
     startRotationX:
       serializedVersion: 2
-      minMaxState: 0
+      minMaxState: 3
```
m4rtint: Did you make any changes to the powerorb blue?

DJSteins: Oh I may have accidentally done that when I was looking at different effects

m4rtint: yeah as long as there's nothing significant change, it's fine

Assets/Script/PowerUp/GravityWellBehaviour.cs
```diff
@@ -100,4 +100,10 @@ private void OnTriggerExit(Collider other)
             m_lengthOfCarsList--;
         }
     }
+
+    
+    protected override float GetYCoord()
+    {
+        return 0.25f;
```
Gursaanj: why do you want such a specific float to be returned here?

DJSteins: It just works to put the indicator right above the ground so it doesn't clip below the map

PR #41, opened 2020-01-12
Map/residential map

m4rtint:
- Residential map
- Added map to main menu flow
- Added in more third party 3D meshes
- Added in Terrain
- Moved a ton of the third party 3d assets into a folder called "Third Party Art"


m4rtint:
[Trigger end of game panel](https://trello.com/c/XM051tTk/354-trigger-end-of-game-panel)

PR #44, opened 2020-01-16
Feature/dissappearing power ups

Gursaanj:
**Deployable based** Powerups will now deactivate after a fixed number of seconds. I did not make this a serialized field since we did not discuss if all the powerups should all last the same amount of time or not. 

Will change if requested (will put as part of the Scriptable Object feature)

Also I did not know what type (or if we even wanted) any animation on the powerups as they disappear, so it is just a blank deactivation for now.


Gursaanj:
[Implement powerup timer for disappearing](https://trello.com/c/a7TOKzCs/340-implement-powerup-timer-for-disappearing)

Assets/Script/PowerUp/BasePowerUp/Deployable.cs
```diff
@@ -23,6 +23,11 @@ public abstract class Deployable : PowerUp
     [SerializeField]
     private float m_initialSpeed = 15.0f;
 
+    //Handling the Deactivation of Powerups
+    private bool m_isWaitingForDeactivation = false;
+    private float m_durationOfPowerup = 3f;
+    private float m_deactivationTimer = 0f;
```
m4rtint: How come you have start and finish? 

Why not just a duration and wait until it hits 0?

Assets/Script/PowerUp/GravityWellBehaviour.cs
```diff
@@ -106,4 +105,9 @@ protected override float GetYCoord()
     {
         return 0.25f;
     }
+
+    private void OnDisable()
```
m4rtint: This is going to break, 

Awake is only called once, 
so once you disable it,  and then you reactivate it, it will still be unsubscribed.

Assets/Script/PowerUp/BasePowerUp/Deployable.cs
```diff
@@ -74,6 +94,10 @@ private void Update()
     {
```
m4rtint: I didn't catch this last time, but this shouldn't be in Update, 
It should be subscribed to the WorldManager FixedUpdate.

Assets/Script/PowerUp/BasePowerUp/Deployable.cs
```diff
@@ -47,6 +52,21 @@ public override void Activate(PlayerController playerController)
...
+
+    public override void Deactivate()
+    {
+        if (m_deactivationTimer < m_durationOfPowerup)
+        {
+            m_deactivationTimer += Time.deltaTime;
+        }
+        else
+        {
+            gameObject.SetActive(false);
```
m4rtint: Place this last, because when you set this to false, all code under this stops working.

Assets/Script/PowerUp/BasePowerUp/Deployable.cs
```diff
@@ -23,6 +23,11 @@ public abstract class Deployable : PowerUp
     [SerializeField]
     private float m_initialSpeed = 15.0f;
 
+    //Handling the Deactivation of Powerups
+    private bool m_isWaitingForDeactivation = false;
```
m4rtint: I honestly don't think we need this. 

Just make the duractionOfPowerUp serialized/readonly
and then have an elapsed time, where it counts down to 0.


then you can use the logic of elapsed time > 0 will mean that the deployable is still active.

Assets/Script/PowerUp/GravityWellBehaviour.cs
Gursaanj: good catch, ill put the subscription to the update funciton in OnEnable

Assets/Script/PowerUp/BasePowerUp/Deployable.cs
Gursaanj: so i tried to that with my first push but I keep getting this error when I try to deploy the Gravity well and wont Instantiate. This only comes up with the GravityWell, must be because that it is also subscribed to the BaseWorldManager.OnFixedUpdate, not sure what causes it though.

I also tried to Subscribe Deployable.cs to OnUpdate, same issue

image

Assets/Script/PowerUp/BasePowerUp/Deployable.cs
Gursaanj: Was wondering a way to cache the value of m_durationOfPowerUp without losing its reference

Assets/Script/PowerUp/BasePowerUp/Deployable.cs
m4rtint: Did you ever google it?

Assets/Script/PowerUp/BasePowerUp/Deployable.cs
m4rtint: Right okay makes sense

Assets/Script/PowerUp/BasePowerUp/Deployable.cs
Gursaanj: yeah, for like 90 mins now, Im kinda stumped to be honest

Gursaanj: will try some work arounds for like another 15 min, and commit/push whatever I have.

Assets/Script/PowerUp/BasePowerUp/Deployable.cs
```diff
@@ -49,6 +55,19 @@ public override void Activate(PlayerController playerController)
...
 
+    public override void Deactivate()
+    {
+        if (m_deactivationTimer < m_durationOfPowerup)
+        {
+            m_deactivationTimer += Time.deltaTime;
+        }
+        else
+        {
+            m_deactivationTimer = 0;
```
m4rtint: You can change a const here?

Assets/Script/PowerUp/BasePowerUp/Deployable.cs
m4rtint: Don't force the workaround, I Googled a bit and I get some answers that it's a Unity Bug. Can't confirm, but if you still can't find out, just leave it alone. And maybe leave some comment about this assertion error above the Update

Assets/Script/Game/DebugInventory.cs
```diff
@@ -14,7 +14,7 @@
 public class DebugInventory : MonoBehaviour
 {
     [SerializeField]
-    private bool m_debugMode = false;
+    private bool m_debugMode = true;
```
mikeslgoh: should we leave this as false

m4rtint: technically yes. but not a problem rn

Assets/Script/PowerUp/BasePowerUp/Deployable.cs
Gursaanj: nah that was a mistake on my part. will revert that now

Assets/Script/PowerUp/BasePowerUp/Deployable.cs
Gursaanj: Done, will add this to a bug/backlog on Trello

PR #45, opened 2020-01-17
Feature/character selection Part 1

m4rtint:
- Joining with controller/keyboard
- Changing Material of mesh
- Confirming Car Color

Just look through some of the cs file, to see if there's any weird code.


Gursaanj:
[Implement new design for player selection](https://trello.com/c/xehnrwL7/353-implement-new-design-for-player-selection)

Assets/Script/UI/MainMenu/CarSelectionPanel.cs
```diff
@@ -91,32 +88,24 @@ public void FullReset()
...
+        AnimateCarSpinning();
     }
-
-    private void Awake()
+    
+    private void AnimateCarSpinning()
     {
-        CheckDependencies();
-        SetAlphaOfCar(1f);
+        CommonAnimations.LoopedRotation(m_carMesh.transform, new Vector3(-1.43f, 360f, 0));
```
Gursaanj: Pretty specific Vector3 , maybe cache it as a private or something at the top so that we can more easily change the values during the playtest or something

Assets/Script/UI/MainMenu/ControllerSelectionPanel.cs
```diff
@@ -278,7 +255,7 @@ private void PCCarCancel()
         }
     }
 
-    private void CarSelected(object control, Player player, int pointer)
+    private void CarSelected(object control, Player player)
```
Gursaanj: does the variable here have to be Object?

Assets/Script/UI/MainMenu/JoinedPlayerPanel.cs
```diff
@@ -14,55 +15,36 @@
 public class JoinedPlayerPanel : MonoBehaviour
 {
     [SerializeField]
+    [Required("Missing Text Mesh Pro Component to change")]
     private TextMeshProUGUI m_text;
 
+    public Sequence InitializeThenShow(float fadeOutTime, float delayedTimeUntilStartOfFade, string s)
```
Gursaanj: better naming convention for the string in this method needed

Assets/Script/UI/MainMenu/JoinedPlayerPanel.cs
```diff
@@ -14,55 +15,36 @@
...
         return CommonAnimations.FadeOutAlpha(GetComponent<CanvasGroup>(), fadeOutTime, delayedTimeUntilStartOfFade);
     }
-
-    /// <summary>
-    /// Changes the joined panel text.
-    /// </summary>
-    /// <param name="s">Text for joined panel</param>
-    public void ChangeJoinedPanelText(string s)
+    
+    private void ChangeJoinedPanelText(string s)
```
Gursaanj: Same here with the naming convention

Assets/Script/UI/MainMenu/ControllerSelectionPanel.cs
m4rtint: yes. keyboard vs controllers

Assets/Script/UI/MainMenu/CarSelectionPanel.cs
m4rtint: I'll make a comment about this value, It'll probably make more sense, rather than making a cached value with a vague name AND commenting on this cache value.

PR #47, opened 2020-01-17
Enhancement/change bouncy tire prefab

mikeslgoh:
- Created a new tire prefab based on the one available in one of the third party assets
- Removed the old bouncy placeholder and copied the components over to the new one

NOTE: need to playtest and rename the powerup when we switch to the new icons

Also added fix to remove the version update for InControl


mikeslgoh:
[Replace bouncy object pickup](https://trello.com/c/YTuCQW2a/348-replace-bouncy-object-pickup)

PR #49, opened 2020-01-18
Enhancement/toolbox pickup

mikeslgoh:
Replaced placeholder cube for the pickup with the toolbox as well as added a rotation on it


mikeslgoh:
[Replace pickup spawner with the toolbox prefab](https://trello.com/c/A1AyBuVS/366-replace-pickup-spawner-with-the-toolbox-prefab)

PR #51, opened 2020-01-18
Input for power up for PC and Controller

m4rtint:


m4rtint:
[Update controller support](https://trello.com/c/4fKXhBSw/372-update-controller-support)

PR #74, opened 2020-01-24
Enhancement/fixed doozy sound effects intergration

Gursaanj:
Realized during GameTech, that the SFX volume wasn't changing with the SFX slider in the OptionPanel. 

Made the necessary changes and also linked the slider to change the volume of the Doozy UI Sound Effects, which uses an audio mixer thus the float needed to be converted into Decibels 

Can't merge to develop, cause playtesting-gametech has yet to be developed


mikeslgoh:
[SFX volume does not reflect on slider](https://trello.com/c/PRc3EiTl/376-sfx-volume-does-not-reflect-on-slider)

PR #78, opened 2020-02-09
New driving controller

DJSteins:
Home-grown-stick-shift


Gursaanj:
[Write our own driving controller script](https://trello.com/c/QkaSOJpV/377-write-our-own-driving-controller-script)

Assets/Script/Game/Car/CarController.cs
```diff
@@ -33,18 +33,15 @@ public class CarController : MonoBehaviour
...
-    private float m_rotate;
-    private float m_rotateTarget;
-    private bool m_nearGround;
-    private bool m_onGround;
     private PlayerInformation m_playerInformation;
     private IInput m_inputManager;
     private PlayerController m_playerController;
 
+    // stores the angle of the steering wheel in degrees
+    float m_steeringAngle;
```
Gursaanj: missing protection level

Assets/Script/Game/Car/CarController.cs
```diff
@@ -195,68 +181,117 @@ private void DependenciesLoad()
...
+        float steeringStep = m_carProperties.SteeringSpeed * Time.deltaTime;
+        if (horizontalMovement != 0)
+        {
+            m_steeringAngle = Mathf.Max(-maxAngle, Mathf.Min(maxAngle, m_steeringAngle + horizontalMovement * steeringStep));
+        }
+        // turn the wheel back to the center
+        else if (m_steeringAngle != 0.0f)
         {
-            CarBreak(m_carProperties.Brake);
+            var previousWheelDirection = m_steeringAngle / Mathf.Abs(m_steeringAngle);
```
Gursaanj: should be float not var I think?

Assets/Script/Game/Car/CarController.cs
```diff
@@ -195,68 +181,117 @@ private void DependenciesLoad()
...
+            m_steeringAngle = m_steeringAngle - previousWheelDirection * steeringStep;
+            if (Mathf.Abs(m_steeringAngle) <= 0.001f)
```
Gursaanj: Not sure if this will help but if youre just looking for a really small float to compare to, you can use mathf.epsilon for a really small float reach towards 0 (but never 0)

Assets/Script/Game/Car/CarController.cs
```diff
@@ -195,68 +181,117 @@ private void DependenciesLoad()
...
+    /// Accelerates the car in the appropriate direction of motion
+    /// </summary>
+    private void AccelerateCar()
+    {
+        float verticalMovement = GetVerticalMovement();
+
+        if (m_steeringAngle != 0)
+        {
+            // do some math to get the radius of a circle tangential to the angle we want to steer to
+            float minRadius = 1.622294f + 1.734175f;
```
Gursaanj: should store float as a serializedField if we plan to play around with this value, or as const if this will never change

Assets/Script/Game/Car/CarController.cs
```diff
@@ -287,33 +322,19 @@ private float GetHorizontalMovement()
...
-    }
-
     private void JumpIfNeeded()
     {
         // Jump
-        if (m_carProperties.JumpAbility && m_nearGround && Input.GetKeyDown(KeyCode.Space))
+        if (m_carProperties.JumpAbility && Input.GetKeyDown(KeyCode.Space))
         {
-            m_sphere.AddForce(Vector3.up * (m_carProperties.JumpForce * 20), ForceMode.Impulse);
+            m_box.AddForce(Vector3.up * (m_carProperties.JumpForce * 20), ForceMode.Impulse);
```
Gursaanj: oooh should we make the float (20f) as a serialized parameter too so that we can test out different forces

Assets/Script/Game/Car/CarController.cs
DJSteins: Epsilon is too small.

Assets/Script/Game/Car/CarController.cs
DJSteins: Jump force is already serialized.

Assets/Script/ScriptableObject/CarProperties.cs
```diff
@@ -48,41 +43,35 @@ public class CarProperties : ScriptableObject
...
-    /// <value>
-    /// The drift.
-    /// </value>
-    public float Drift => m_drift;
+    /// <summary> 
+    /// Gets the drag which is a value from 0 to 1 
+    /// Heavier cars have less drag and higher handling gives you better drag
+    ///</summary>
+    /// <value> The drag </value>
+    public float Drag => ((1.0f - m_mass / 2.0f) + m_handling) / 2.0f;
```
m4rtint: Sometimes I wonder if this should be a property or a method.

DJSteins: Could be considered a "lambda" method?

m4rtint: it's not really used like a lambda method is it? 
either it works.

Assets/Script/Game/Car/CarController.cs
```diff
@@ -67,6 +78,7 @@ public class CarController : MonoBehaviour
             value = m_playerController;
```
m4rtint: Swap this, needs to be fixed.

Assets/Script/Game/Car/CarController.cs
```diff
@@ -122,26 +118,30 @@ public class CarController : MonoBehaviour
     /// </summary>
     public Player Player => m_playerController.Player;
 
+    /// <summary>
+    /// Gets the car rotation around the y-axis
+    /// </summary>
+    public float Rotation => transform.eulerAngles.y;
+
+    public bool OnGround => transform.position.y <= m_ModelYOffset && transform.position.y >= 0;
```
m4rtint: Make this into a method.

Assets/Script/Game/Car/CarController.cs
```diff
@@ -189,198 +190,279 @@ private void OnDestroy()
...
+        if (m_steeringAngle != 0)
         {
-            CarBreak(m_carProperties.Brake);
+            float radius = GetTurnRadius();
+            float omega = GetAngularTurnRate(radius);
+            float turnDirection = GetTurnDirection();
+            // specify the center point for the circle we will rotate around
+            Vector3 centerOfRotation = GetCenterOfRotation(radius, omega, turnDirection);
+            // rotate the car
+            m_box?.transform.RotateAround(centerOfRotation, Vector3.up, omega * Time.deltaTime * GetVerticalMovement());
```
m4rtint: is there any point the m_box is going to be null?

Assets/Script/ScriptableObject/CarProperties.cs
m4rtint: Can you make this into a method.

m4rtint:
With the editor crashing, i left a few comments 
1) One big problem
2) the rest, convert from property to method. 

I don't know if these will fix the accessing bad memory issue, but i'm trying to cover all bases just incase

Assets/Script/Game/Car/CarController.cs
m4rtint: this might actually be the thing that's causing crashes. Might need to hot fix this.

Assets/Script/Game/Car/CarController.cs
DJSteins: I thought it had something to do with the crashes. Guess not.

PR #79, opened 2020-02-22
Update Doozy with base layer loading screen

m4rtint:
Assets/Script/Managers/AudioManager.cs
Assets/Script/Managers/PlayerManager.cs

Only files I've changed for code.


Assets/Script/Managers/PlayerManager.cs
```diff
@@ -143,7 +143,7 @@ private void InitializedCarSpawnPoints()
 
     private void CreatePlayer(PlayerInformation playerInfo, Player playerNumber)
     {
-        PlayerController playerController = Instantiate(m_playerController_Template.gameObject).GetComponent<PlayerController>();
+        PlayerController playerController = Instantiate(m_playerController_Template.gameObject, transform).GetComponent<PlayerController>();
```
Gursaanj: This makes the playerController GameObject a child of the PlayManager right? Any reason for doing this. or is it just for neatness sake?

m4rtint: Yes. 
So the flow currently is
MainMenu Loaded + Active -> Generic UI + Map Loaded/Inactive

Generic UI + Map Loaded/Active
Players are created in the Main Menu Scene (it's still the first scene)

Unload Main Menu -> Players get unloaded

Assets/Doozy/Engine/Scripts/Coroutiner.cs
```diff
@@ -45,6 +45,14 @@ public static Coroutiner Instance
 
         #region Unity Methods
 
+        #if UNITY_2019_3_OR_NEWER
+        [RuntimeInitializeOnLoadMethod(RuntimeInitializeLoadType.SubsystemRegistration)]
+        private static void RunOnStart()
```
wbeltran777: What's the purpose of having this function in three different classes? If it's private static does it mean that they're all modifying the same ApplicationIsQuitting bool?

m4rtint: I did not write this.

wbeltran777: Alrighty! Then everything looks great hahahah

PR #85, opened 2020-03-03
Feauture/forward button

m4rtint:


m4rtint:
[Implement new UI for back and confirm buttons](https://trello.com/c/ISb5xTHL/401-implement-new-ui-for-back-and-confirm-buttons)

mikeslgoh:
Is this branch pointing to the wrong branch to be merged into @m4rtint

Assets/Script/UI/Loading/LoadingMenuView.cs
```diff
@@ -0,0 +1,70 @@
...
+        private void Start()
+        {
+            BindingDelegates();
+            m_hasFinishedLoading = false;
+            m_forwardButton.HideButton();
+        }
+
+        private void BindingDelegates()
+        {
+            m_pointer.OnCompleteLoading += OnCompleteLoadingAnimation;
```
Gursaanj: This event never seems to be unsubscribed in this case

Assets/Script/UI/MainMenu/HoldButton/HoldButton.cs
```diff
@@ -31,33 +26,52 @@ public abstract class HoldButton : MonoBehaviour
...
+            m_canHold = true;
             ResetTimer();
             ResetLoadingButton();
         }
     }
 
+    /// <summary>
+    /// Set to hold button to go back/forward.
+    /// </summary>
+    public void SetToHoldDuration()
```
Gursaanj: More of a question than comment. I know that Public Methods should be before Private ones, But should Public methods be placed before protected ones too??

Cause I think the convention is Public -> Protected -> Private

m4rtint: uhhh..house rules. no clue. 
you can check with stylecop .

but yes I think you are correct

Gursaanj: According to stackOverflow

According to the StyleCop Rules Documentation the ordering is as follows.

**Within a class, struct or interface: (SA1201 and SA1203)**

Constant Fields
Fields
Constructors
Finalizers (Destructors)
Delegates
Events
Enums
Interfaces (interface implementations)
Properties
Indexers
Methods
Structs
Classes


**Within each of these groups order by access: (SA1202)**

public
internal
protected internal
protected
private

**Within each of the access groups, order by static, then non-static: (SA1204)**

static
non-static

**Within each of the static/non-static groups of fields, order by readonly, then non-readonly : (SA1214 and SA1215)**

readonly
non-readonly


An unrolled list is 130 lines long, so I won't unroll it here. The methods part unrolled is:

public static methods
public methods
internal static methods
internal methods
protected internal static methods
protected internal methods
protected static methods
protected methods
private static methods
private methods

Gursaanj: SA1201 Seems like we would have to change too much (and honestly way too much of a pain) but the rest seems like we can follow without much refactoring!

We should bring it up next coding meeting

Assets/Script/UI/Loading/LoadingMenuView.cs
m4rtint: Moved this to OnEnable and OnDisable

Assets/Script/UI/MainMenu/HoldButton/HoldButton.cs
m4rtint: You would think so, but i don't think anyone would focus on this. 

We still have testing in the back log. 

I don't think we need to focus on this.

mikeslgoh:
@Gursaanj could you help to review and maybe approve this? Thanks

PR #90, opened 2020-03-11
Better driving controller

DJSteins:
Adds ramp support and modifies the input from keys to steering to work how it did with the old driving controller.


Assets/Script/Game/Car/CarController.cs
```diff
@@ -343,13 +371,26 @@ private void AccelerateCar()
         m_box.AddForce(acceleration * Time.deltaTime, ForceMode.VelocityChange);
     }
 
+    private void RotateToGround()
+    {
+        // reference: https://answers.unity.com/questions/168097/orient-vehicle-to-ground-normal.html?_ga=2.154419966.778972075.1583965656-1799704818.1555470483
+        RaycastHit hit;
+        if (Physics.Raycast(m_box.transform.position, -m_curNormal, out hit))
+        {
+            m_curNormal = Vector3.Lerp(m_curNormal, hit.normal, 4 * Time.deltaTime);
```
m4rtint: The last time i saw you testing, was when you rotated the floor.  (So the Rotation of the floor is not Quart.identity)
Have you tried testing with an actual ramp object where the mesh is zeroed out? Does that give you the correct normal?

Assets/Script/Game/Car/CarController.cs
```diff
@@ -273,17 +277,41 @@ private Vector3 GetCenterOfRotation(float radius, float omega, float turnDirecti
         return wheelPosition + radius * (Quaternion.AngleAxis(turnDirection * 90.0f, Vector3.up) * forward);
     }
 
+    private float GetCameraTurnModifier()
+    {
+        if (!m_isHolding)
+        {
+            return 1;
+        }
+        float dt = Time.time - m_horizontalHeldDown;
```
m4rtint: what is the horizontalHeldDown for? why is it related to time?

Assets/Script/Game/Car/CarController.cs
```diff
@@ -55,6 +55,10 @@ public class CarController : MonoBehaviour
     /// Use this timer to override the movement of the car
     /// </summary>
     private float m_movementOverrideTimer;
+    Vector3 m_curNormal = Vector3.up;
+    float m_horizontalHeldDown = 0.0f;
+    bool m_isHolding = false;
+    float m_lastSignal = 1;
```
Gursaanj: Missing Protection Levels

Assets/Script/Game/Car/CarController.cs
DJSteins: Actual ramp object? If you look at the video I put in the coding channel the ramp there is a plane with a box collider on it and it works fine on that ramp.

Assets/Script/Game/Car/CarController.cs
DJSteins: It's for checking how long the right or left directional input is held down. This is to handle the case where the driver is holding down left or right while the direction of the car changes from facing towards the camera to facing away from the camera or vise-versa.

Assets/Script/Game/Car/CarController.cs
m4rtint: Mmm those ramps are rotate to a certain degrees.
If you were to use ramps which are rotation 0,0,0, would you get the same outcome?

m4rtint:
PLease merge from develop

Assets/Script/Game/Car/CarController.cs
m4rtint: Tested this out, it works np

m4rtint:
Bug: I'm not sure if it's related to the ramps. 
When you hold down forward and press left/right/ over the ramps?? 
and then u switch directions, it will go the wrong direction.

Hold forward + hold right
go over ramp
Keep holding forward + hold Left.

Car still goes right. 

Hard to pin point  exact repo steps, but easy to do it randomly

mikeslgoh:
[Refactor car controller to test with a ramp](https://trello.com/c/KZNUQQls/411-refactor-car-controller-to-test-with-a-ramp)

PR #91, opened 2020-03-15
Bugs/refactor stuff

m4rtint:
This one is a bit insane. 
But it's 
- Code adding namespaces
- Code adding = null
- Organizing files
- Changed - StoredString -> Data, Placed all stored strings into their respective structs


m4rtint:
[Refactor Project](https://trello.com/c/2XEtmRTK/422-refactor-project)

Assets/_PartyParkade/Script/Data/CommonAnimations.cs
```diff
@@ -0,0 +1,82 @@
+﻿//-----------------------------------------------------------------------
+// <copyright file="CommonAnimations.cs" company="PERIGON GAMES">
+//     Copyright (c) PERIGON GAMES. 2019 All Rights Reserved
+// </copyright>
+//-----------------------------------------------------------------------
+
+using DG.Tweening;
+using UnityEngine;
+
+namespace _PartyParkade.Script.Data
```
Gursaanj: namespaces having an "_" seem a bit off to me!! Is this regular convention?

m4rtint: You're supposed to name space the folders

mikeslgoh:
Is it possible to narrow down the relevant files that we should be looking at? Just because I don't think we will be able to navigate through 4000 file changes

PR #94, opened 2020-03-17
Set instances to null

m4rtint:


m4rtint:
[Refactor Project](https://trello.com/c/2XEtmRTK/422-refactor-project)

PR #107, opened 2020-03-22
Bug/Falling Objects

DJSteins:
Moved dumpsters and trucks down on eight map.


mikeslgoh:
[Physics of all objects drops when game is started](https://trello.com/c/reXSN7F1/423-physics-of-all-objects-drops-when-game-is-started)

PR #112, opened 2020-03-23
Fix Failing Parking Indicator Test

DJSteins:
Modified the expected result of one of the parking indicator tests. Since the method calculates the angle about Vector.up, giving Vector.up as a direction should return a zero vector.


DJSteins:
[Parking Indicator test failing](https://trello.com/c/1a377PlM/441-parking-indicator-test-failing)

PR #114, opened 2020-03-29
Generic Object Pooler

m4rtint:
- This is mainly so that the menu has object pooler swap between car selection


Assets/_PartyParkade/Script/Data/ObjectPooler/ObjectPool.cs
```diff
@@ -0,0 +1,73 @@
+﻿using System;
```
Gursaanj: Missing Header

Assets/_PartyParkade/Script/Data/ObjectPooler/ObjectPool.cs
```diff
@@ -0,0 +1,73 @@
...
+
+            return spawnedItem;
+        }
+    }
+
+    public struct PoolerObject<T> where T : Component
+    {
+        public int m_tag;
+        public T m_pooledObject;
+        public int m_size;
```
Gursaanj: if these are public, I dont think the they need "m_"

Assets/_PartyParkade/Script/Data/ObjectPooler/ObjectPool.cs
```diff
@@ -0,0 +1,73 @@
...
+            if (!m_poolDictionary.ContainsKey(poolTag))
+            {
+                Debug.LogWarning(message: $"{tag} does not exist");
+                return null;
+            }
+
+            T pooledObj = m_poolDictionary[poolTag].Dequeue();
+            m_poolDictionary[poolTag].Enqueue(pooledObj);
+            
+            return pooledObj;
```
Gursaanj: you can simplify this contains key check with m_poolDicitonary.TryGetValue(poolTag, out value)

https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.dictionary-2.trygetvalue?view=netframework-4.8

Assets/_PartyParkade/Script/Data/ObjectPooler/ObjectPool.cs
```diff
@@ -0,0 +1,101 @@
...
+            if (!m_poolDictionary.ContainsKey(poolKey))
+            {
+                Debug.LogWarning(message: $"{tag} does not exist");
+                return null;
+            }
+
+            T pooledObj = m_poolDictionary[poolKey].Dequeue();
+            m_poolDictionary[poolKey].Enqueue(pooledObj);
+            
+            return pooledObj;
```
Gursaanj: to simply the containskey check, you could use m_pooldictionary.TryGetValue(poolkey, out value)

https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.dictionary-2.trygetvalue?view=netframework-4.8

Assets/_PartyParkade/Script/Data/ObjectPooler/ObjectPool.cs
```diff
@@ -39,23 +51,35 @@ public void Initialize(List<PoolerObject<T>> differentObjectsToSpawn)
...
+        /// Gets an object from with a stored key
+        /// </summary>
+        /// <param name="poolKey">key</param>
+        /// <returns>Pooled Object</returns>
+        public T PopPooledObject(int poolKey)
         {
-            if (!m_poolDictionary.ContainsKey(poolTag))
+            if (!m_poolDictionary.ContainsKey(poolKey))
             {
                 Debug.LogWarning(message: $"{tag} does not exist");
```
Gursaanj: should tag be poolkey in the message?

Assets/_PartyParkade/Script/Data/ObjectPooler/ObjectPool.cs
m4rtint: I don't really see the difference except that I will need to cache the Queue.

m4rtint: This would basically 

var tempQCar
if (m_poolDictionary.TryGetValue(poolkey out tempQCar)
{
    return null
}

T pooledObj = tempQCar.Dequeue()
tempQCar.Enqueue(pooledObj)


Thoughts?

Gursaanj: yeah it would save the unnecessary lookup for pooledObj and tempQCar.enque

but nothing too crazy

Assets/_PartyParkade/Script/Data/ObjectPooler/ObjectPool.cs
```diff
@@ -0,0 +1,103 @@
...
+            }
+        }
+
+        #endregion
+        
+        /// <summary>
+        /// Initialize the pooled of objects
+        /// </summary>
+        /// <param name="differentObjectsToSpawn"> List of different objects to spawn</param>
+        public void Initialize(List<PoolerObject<T>> differentObjectsToSpawn)
```
wbeltran777: Since we need to pass a list of pooler objects, would it be useful to have a function that given an arbitrary list of T components, outputs a list of PoolerObject<T>? 
I bring it up cause if so then we'd want it to live in this function, otherwise we'd be creating lists of PoolerObjects elsewhere and it could get messy.

m4rtint: Good point

wbeltran777: live in this class***** 
not function!

m4rtint: Made a create function

m4rtint:
[Generic Object pooler](https://trello.com/c/l0Qv1V6S/444-generic-object-pooler)

PR #118, opened 2020-04-05
Feature/powerup collisions

DJSteins:
Power-ups now bounce back when they collide with something that isn't the car, road, or parking spot. This is based on game object tags and requires that every deployable power-up has a collider that is a trigger and a kinematic rigidbody.


mikeslgoh:
[Deployment of Poweups collide through objects](https://trello.com/c/xKIH6fx7/371-deployment-of-poweups-collide-through-objects)

mikeslgoh:
Does this PR include the change that everyone will be affected by powerups?

DJSteins:
@mikeslgoh that was in the gravity well PR and it was only gravity well that wasn't affecting every player.

Assets/_PartyParkade/Script/PowerUp/BasePowerUp/Deployable.cs
```diff
@@ -114,4 +116,30 @@ protected virtual float GetYCoord()
...
+    /// </summary>
+    /// <param name="other">The collider that collided with this.</param>
+    private void OnTriggerEnter(Collider other)
+    {
+        if (other.gameObject.CompareTag(Tags.PlayerCar))
+        {
+            HandleTrigger(other);
+        }
+        else if (
+            !other.gameObject.CompareTag("RoadConnection") &&
```
m4rtint, line 131: Place these data strings into the Data file

DJSteins: Even better would be if we had a general tag/identifier for all things that the power-up could/should bounce off of

m4rtint: the problem with that is , you can only tag each item once.

DJSteins: :(

PR #126, opened 2020-04-12
Feature/polish boost

DJSteins:
- Added particle effects to the boost power-up
- Changed boost logic to be a function of boost percent that decrements based on a boost percent step
- Modified car controller turning to use AddForceAtPosition at each of the front wheel positions


DJSteins:
[Polish Boost](https://trello.com/c/1Wj4lpJN/457-polish-boost)

Assets/_PartyParkade/Script/Game/Car/CarController.cs
```diff
@@ -41,12 +43,19 @@ public class CarController : MonoBehaviour
...
+        [SerializeField]
+        private GameObject m_boostTrails = null;
 
         private IInput m_inputManager = null;
         private PlayerController m_playerController = null;
 
         private float m_previousVerticalMovement;
 
+        // 0 <= m_boostPercent <= 100 is used to track how much boost is left
+        private float m_boostPercent;
```
m4rtint, line 55: Do you think this could be a variable depending on the vehicle?

m4rtint:
Just a suggestion, you could make a separate  script dedicated to boost, which is attached to the same object as CarController. 

So that not everything will be in carController, and the car controller can reference the other boost script, and the logic could be separated.

Assets/_PartyParkade/Script/PowerUp/Boost.cs
```diff
@@ -10,7 +10,9 @@
 
 public class Boost : Miscellaneous
 {
-    [SerializeField]private float boostPower = 700f;
+    [SerializeField]
+    private float boostPower = 1f;
```
wbeltran777, line 14: If we decide to have different boost powers for different car types, it could be good to move this boostPower field to the Car Controller. Not a huge issue at the moment though

Assets/_PartyParkade/Script/Game/Car/CarController.cs
```diff
@@ -153,11 +162,19 @@ public void Initialize(Vector3 position, PlayerInformation information, IInput i
...
 
+        /// <summary>
+        /// Activates boost for this car.
+        /// </summary>
+        /// <param name="boostPower">The boost power</param>
         public void BoostActivate(float boostPower)
         {
-            m_carRigidBody.AddForce(-m_body.forward * boostPower, ForceMode.Impulse);
+            EmitBoost(true);
+            m_boostPercent = 100;
```
wbeltran777, line 176: Nit pick, we could save 100 as a const float

Assets/_PartyParkade/Script/Game/Car/CarController.cs
```diff
@@ -229,84 +246,43 @@ private void CarControllerOnFixedUpdate()
...
-                float radius = GetTurnRadius();
-                float omega = GetAngularTurnRate(radius);
-                float turnDirection = GetTurnDirection();
-                float reverseModifier = GetReverseTurnModifier();
-                // specify the center point for the circle we will rotate around
-                Vector3 centerOfRotation = GetCenterOfRotation(radius, omega, turnDirection);
-                // rotate the car
-                m_carRigidBody.transform.RotateAround(centerOfRotation, Vector3.up, omega * Time.deltaTime * reverseModifier);
-            }
+            return m_steeringAngle / Mathf.Abs(m_steeringAngle);
```
Gursaanj, line 285: hmm I wonder if it would be safer to check if m_steeringAngle != 0, no matter how unlikely before returning this (instead of assuming). Cause Unity will crash with memory leak otherwise

Assets/_PartyParkade/Script/Game/Car/CarController.cs
```diff
@@ -229,84 +246,43 @@ private void CarControllerOnFixedUpdate()
...
-        /// </summary>
-        private Vector3 GetCenterOfRotation(float radius, float omega, float turnDirection)
+        
+        private void EmitBoost(bool emit)
         {
-            Vector3 forward = transform.forward.normalized;
-            forward.y = 0.0f;
-            Vector3 wheelPosition = omega > 0 ? m_wheelFrontRight.position : m_wheelFrontLeft.position; 
-            return wheelPosition + radius * (Quaternion.AngleAxis(turnDirection * 90.0f, Vector3.up) * forward);
+            foreach (ParticleSystem particles in m_boostTrails.GetComponentsInChildren<ParticleSystem>())
```
Gursaanj, line 256: since EmitBoost is called on Update, using GetComponentsInChildren can be costly if not necessary. Maybe let an array = m_boostTrails.GetCompnentInChildren<ParticleSystem>() at the start and just iterate over that

Assets/_PartyParkade/Script/Game/Car/CarController.cs
DJSteins: I think boost percent should be between 0 to 100 because it is a percent, but the m_boostReductionPerStep could be a vehicle property.

Assets/_PartyParkade/Script/Game/Car/CarController.cs
DJSteins: I already do that check before calling the function.

mikeslgoh:
@m4rtint @wbeltran777 @Gursaanj please help to review it again and then add the autosquash label

PR #134, opened 2020-04-19
Update Nav Section to be disabled if not needed

m4rtint:


Assets/_PartyParkade/TrafficSystem/NavSection.cs
```diff
@@ -23,6 +23,19 @@ public virtual void Start()
         {
             foreach (NavConnection connection in connections)
                 connection.navSection = this;
```
Gursaanj, line 25: could you add CBs here , for consistency

m4rtint: Added

PR #142, opened 2020-04-26
Switched the shader from the Dust material to avoid crash

mikeslgoh:
For some reason after testing it, the error we found is still being thrown but the game does not crash anymore after switching the material.

NOTE: Unity is planning to fix this but only in 2020 edition
https://issuetracker.unity3d.com/issues/task-dot-rasterdata-dot-vertexbuffer-equals-equals-null-errors-when-using-a-particle-system


mikeslgoh:
Please help to playtest in your own time as well to make sure the crash doesn't occur on your end anymore.

mikeslgoh:
[Game crashes due to particle effects](https://trello.com/c/PV6wMJHb/481-game-crashes-due-to-particle-effects)

PR #146, opened 2020-04-29
Changes made during playtesting on 27th April

mikeslgoh:
- Added winterland to the build settings as a selectable map
- Also updated the DoozyUI for it
- Added switch player powerup as a debug option


m4rtint:
Do you remmeber what you changed in the player prefab? 
I think it's just removing stuf?

mikeslgoh:
> Do you remmeber what you changed in the player prefab?
> I think it's just removing stuf?

Yeah I think it's just removing stuff so I'll just pull from develop first

PR #148, opened 2020-04-30
Refactor/use car stats for car controller

mikeslgoh:
- Refactored CarController to take CarStats as a parameter
- Created an interface ICarStats for the scriptable object, CarStats to conform too


Assets/_PartyParkade/Script/Game/Car/CarController.cs
```diff
@@ -35,11 +35,11 @@ public class CarController
...
         /// <param name="drag">the amount of drag for the car</param>
-        public CarController(float force, float mass, float drag)
+        public CarController(ICarStats carStats)
         {
-            m_force = force;
-            m_mass = mass;
-            m_drag = drag;
+            m_force = carStats.GetForce();
+            m_mass = carStats.GetMass();
+            m_drag = carStats.GetDrag();
```
m4rtint, line 42: You need to store the carStat as a reference, instead of m_force,m_mass...

so delete m_force,m_mass,m_drag

and store ICarStats as a variable 

and then everything underneath uses 
carStat.getForce()
carStats.getMass()

since this will be reference the Scriptable object, instead of the variables that set once in the constructor.

Assets/_PartyParkade/Script/Game/Car/ICarStats.cs
```diff
@@ -0,0 +1,10 @@
+﻿using System.Collections;
```
m4rtint, line 1: @Gursaanj will complain about Headers and comments here

mikeslgoh: Yeah I should probably do that

mikeslgoh:
[Update CarController to use CarStats](https://trello.com/c/I0eMXUrk/490-update-carcontroller-to-use-carstats)

Assets/_PartyParkade/Script/Game/Car/ICarStats.cs
m4rtint: I meant like, new scripts/files you create, can probably be 2020. but it's fine

Assets/_PartyParkade/Tests/CarController_Tests.cs
```diff
@@ -6,6 +6,22 @@
```
Gursaanj, line 6: Missing Header

PR #156, opened 2020-05-05
Feature/occlusion fix

wbeltran777:
- Added Custom Forward Renderer. URP uses the Forward Renderer to render Scenes. Creating a custom one lets us go into each step of the forward renderer and perform operations. In this case, we're using it to render objects of the CullSphere layer infront of those in the Obstacles layer. 
- Removed all occlusion shaders as the occlusion is now done by the forward renderer. 
- Added the tags and layers for obstacles in each scene. 
- Changed the shader that the simpleTown.mat refers to as it used to be one of the now removed shaders.


mikeslgoh:
[Rework culling shader](https://trello.com/c/BPhERjUu/474-rework-culling-shader)

PR #158, opened 2020-05-05
Playtest/car attributes

mikeslgoh:
Includes some bug fixes as well


m4rtint:
Camera still has empty scripts in it on each map

PR #159, opened 2020-05-05
Feature/parking spot fx

m4rtint:
I just replaced the parking spots on the map


m4rtint:
Oh and the parking spot detector is like super big now.

m4rtint:
REsidential parking will be broken, i did not fix that, i can probably fix it in another PR

@wbeltran777 and you would have to remove and replace your spots too. 

i did not just change the parking objects, i had to remove and replace them all with the parking prefab.

PR #161, opened 2020-05-07
Added Parking to Residential

m4rtint:
image


m4rtint:
[Use particle effects for new parking spot glow](https://trello.com/c/d10mb3DA/450-use-particle-effects-for-new-parking-spot-glow)

PR #167, opened 2020-05-09
Feature/Fixed Turning After Releasing Accelerate

DJSteins:
Adds a "spin-down" model to the acceleration so that it does not immediately cutoff. This allows players to continue turning the car for a bit even after they release acceleration.


mikeslgoh:
[CarController tweaks](https://trello.com/c/UN3eb2l2/503-carcontroller-tweaks)

PR #171, opened 2020-05-11
Feature/parking debugger

Gursaanj:
Created  a **parking spot debugger tool**, that lets us choose which parking spot we want to set as the current destination, a maybe useful tool to test out new parking spots. 

You can't see it clearly in the video but here are the details

- The window only works in Play mode, as soon as you go back to editor mode, the window closes
- The window works between scenes, meaning you can carry the window from one map to another in one play session and it still work 
- When you select a parking Spot, it highlights the position of the parking spot in the Scene view, so that you can see where the parking spot is before you decide to change it 

parking spot debugger


Gursaanj:
[Create Parking Spot Debugging Tool](https://trello.com/c/7NxWs3cc/516-create-parking-spot-debugging-tool)

Assets/Editor/PartyParkadeParkingDebugger.cs
```diff
@@ -0,0 +1,110 @@
...
+    [MenuItem("Window/Party Parkade/Parking Spot Debugger")]
+    private static void OpenWindow()
+    {
+        m_window = (PartyParkadeParkingDebugger)GetWindow(typeof(PartyParkadeParkingDebugger));
+        m_window.minSize = new Vector2(375, 710);
+        m_window.autoRepaintOnSceneChange = true;
+        m_window.Show();
+
+        // To avoid complications with using the window between edit and PlayMode
+        EditorApplication.playModeStateChanged += delegate { m_window.Close(); };
```
m4rtint, line 31: If you keep opening the window, wouldn't this keep attaching to the delegate, creating memory leak?

Assets/Editor/PartyParkadeParkingDebugger.cs
```diff
@@ -0,0 +1,110 @@
...
+
+            for (int i = 0; i < m_parkingSpots.Length; i++)
+            {
+                parkingSpotNames[i] = m_parkingSpots[i].name;
+            }
+
+            m_currentParkingSpotIndex = GUILayout.SelectionGrid(m_currentParkingSpotIndex, parkingSpotNames, 3, GUILayout.Height(350f));
+            m_currentParkingSpot = m_parkingSpots[m_currentParkingSpotIndex];
+            Selection.activeGameObject = m_currentParkingSpot;
+            //SceneView.FrameLastActiveSceneView();
```
m4rtint, line 80: Any news on this problem??

Assets/_PartyParkade/Script/Managers/ParkingManager.cs
```diff
@@ -30,6 +30,22 @@ public class ParkingManager : MonoBehaviour
     /// <value>The current target.</value>
     private GameObject m_currentTarget = null;
 
+    public GameObject[] Targets
```
m4rtint, line 33: Targets => m_targets is less lines of code

Assets/Editor/PartyParkadeParkingDebugger.cs
Gursaanj: Good point, the example I saw for this never unsubscribed so I thought it was okay but, no harm unsubscribing on destroy

Assets/Editor/PartyParkadeParkingDebugger.cs
Gursaanj: Not yet, Like it works but it sometimes crashes the editor, so im keeping it here for now and maybe when I figure it Ill make it work!!

Assets/Editor/PartyParkadeParkingDebugger.cs
```diff
@@ -0,0 +1,124 @@
...
+    {
+        m_parkingManager = FindObjectOfType<ParkingManager>();
+
+        if (m_parkingManager != null)
+        {
+            m_parkingSpots = m_parkingManager.Targets;
+
+            if (m_parkingSpots != null)
+            {
+                m_currentParkingSpot = m_parkingManager.CurrentTarget;
```
m4rtint, line 57: not sure if this is a logic mix up, but why use m_parkingManager here, when you already checked the null value in the if statement above

m4rtint: Should also test what happens if there are no parking spots on the map at all.

m4rtint:
Do you know why Assets/SoundManagerPro/Scripts/Extensions/DFGUI_SMPRO_INTEGRATION.unitypackage.meta
is added?

m4rtint:
image
If there are too many parking slots, UI is messed up, maybe look into scroll view

m4rtint:
[Party-Parkade - ResidentialMap - PC, Mac & Linux Standalone - Unity 2019.3.0f6 Personal [PREVIEW PACKAGES IN USE] _DX11_ 2020-05-11 18-22-56.mp4.zip](https://github.com/PerigonGames/Party-Parkade/files/4612873/Party-Parkade.-.ResidentialMap.-.PC.Mac.Linux.Standalone.-.Unity.2019.3.0f6.Personal.PREVIEW.PACKAGES.IN.USE._DX11_.2020-05-11.18-22-56.mp4.zip)

https://1drv.ms/u/s!ArQDn0JFxY433ydcU0heNYe-mEiA?e=5PJsLB

- Particle Effects don't move with the spot 
- Residential Map has one broken spot

Gursaanj:
So this has been updated to allow for a dynamic scroll view of parking spots, so if there are 3 or 300 parking spots in a map, the scroll view should be able to handle everything cleanly

Fixed the sceneview bug, so that when you select a parking spot, it will move the scenview camera accordingly  to better visualize what parking spot you are looking at

Fixed the bug  where, the panel will update automatically when you park the car in a spot and a new spot has been chosen

PR #172, opened 2020-05-11
Made car wind down stronger after releasing accelerate

DJSteins:


mikeslgoh:
[CarController tweaks](https://trello.com/c/UN3eb2l2/503-carcontroller-tweaks)

PR #174, opened 2020-05-12
Bugfix/parking lines layers

wbeltran777:
So turns out this issue is pretty common
https://forum.unity.com/threads/shader-material-priority-and-renderqueue-for-transparent-materials.826305/

Basically, stuff is rendered in a certain order. Order is determined by a material's render queue value. UI, by default, has a render queue of 3000.
Transparent objects, like the parking line, have a render queue default of 3050. 
For some reason URP only allows you to move this value by a margin of 50. So we can set transparent objects to anywhere between 3000 and 3100. 

So best we can do is set the render queue to 3000. May cause trouble in the future but for now the parking lines won't overlap with the UI.


mikeslgoh:
[Redo Sorting layer for ingame](https://trello.com/c/Rh3y6Lrk/458-redo-sorting-layer-for-ingame)

PR #175, opened 2020-05-12
Bugfix/culling setup

wbeltran777:
Cullsphere wasn't in the correct layer and the sphere mesh wasn't enabled. 

Cullsphere gameobject is now in the Cullsphere layer. Mesh is enabled.

This should fix the culling


mikeslgoh:
[Culling bug in develop](https://trello.com/c/GQKZ9dwO/524-culling-bug-in-develop)

PR #176, opened 2020-05-13
Feature/player selection testing

DJSteins:
Adds player selection testing on the CarSelectionViewModel. Uses a test implementation of the model rather than using PC/Controller implementations of it to avoid reference issues with InControl and/or InputDevice.


mikeslgoh:
[Take a look at PlayerSelection for testing](https://trello.com/c/oN0XWl8S/509-take-a-look-at-playerselection-for-testing)

Assets/_PartyParkade/Tests/CarSelectionViewModel_Tests.cs
```diff
@@ -0,0 +1,222 @@
...
+
+    public class CarSelectionViewModel_Tests
+    {
+        private CarSelections carSelections;
+        private CarHats carHats;
+
+        [SetUp]
+        public void Setup()
+        {
+            carSelections = AssetDatabase.LoadAssetAtPath<CarSelections>("Assets/_PartyParkade/Script/ScriptableObject/ListOfCarsToSelect.asset");
```
m4rtint, line 77: Hmmm how come you need to load the asset?
and if we need to reference with hardcoded path, it's going to cause problems if anything were to be moved or renamed

DJSteins: I don't know how else to get the asset...

m4rtint: So to test this, you'll need to refactor the scriptable object a bit. 

Make CarSelection and CarHats conform to an interface. - Like how CarStats conform to an interface right now. 

and then pass in the interface into this ViewModel and Mock the interface.

PR #179, opened 2020-05-13
Ui/loading screen

mikeslgoh:
The assets were in the project but the scene changes weren't merged in for some reason


m4rtint:
image


This is 1440p

2080x1440p

Confirm this is intended? - Controller touching the left edge

m4rtint:
The title looks fine, without using anchors and stuff. 

Even changed between resolution and it looks fine. Not sure if anyone wants to refactor it later on.

mikeslgoh:
> The title looks fine, without using anchors and stuff.
> 
> Even changed between resolution and it looks fine. Not sure if anyone wants to refactor it later on.

I think we can always go through the UI and double check with different resolutions again since we have QA now too

PR #183, opened 2020-05-14
Feature/display cars end game

wbeltran777:
Two things that I had to tweak!  

- Add DontDestroyOnLoad to ModelCollectionViewController so it persists into the stage and we can use the same images we use in the stage select. It has a singleton-ish pattern where if we go back to the main menu then we destroy the current ModelCollectionViewController. That way the cars go back to their grayed out defaults. 
- Removed the "DOTween.Kill(targetOrId: RotationId);" as otherwise the cars stop spinning when going back to the main menu.


Assets/_PartyParkade/Script/UI/MainMenu/PlayerSelection/3DModelView/CarSelectionMeshViewController.cs
```diff
@@ -122,11 +122,6 @@ private void RotateThisObject()
                 .SetId(RotationId);
         }
```
m4rtint, line 124: Why'd you remove the kill?

Assets/_PartyParkade/Script/UI/MainMenu/PlayerSelection/3DModelView/ModelCollectionViewController.cs
```diff
@@ -20,9 +20,21 @@ public class ModelCollectionViewController : MonoBehaviour
     {
         private CarSelectionMeshViewController[] m_meshControllers = null;
         public CarSelectionMeshViewController[] MeshViewController => m_meshControllers;
+
+        private static ModelCollectionViewController instance = null;
         
         private void Awake()
         {
+            if (instance != null && instance != this)
```
m4rtint, line 28: I'm usually suspicous of any long term problems for singletons,  especially something that isn't a utility/manager.

Anyway you can just put 
instance = this
outside the if statement since it'll always be executed no matter what

Assets/_PartyParkade/Scenes/DoozyMainMenu.unity
```diff
@@ -254,7 +254,7 @@ MonoBehaviour:
     spaceCount: 7
```
m4rtint, line 254: What did you change in the Main Menu?

Assets/_PartyParkade/Script/UI/MainMenu/PlayerSelection/3DModelView/CarSelectionMeshViewController.cs
m4rtint: Oh I just read ur explanation. 

Check OnEnable and OnDisable, don't remove the Kill cuz there might be memory leaks if you don't kill it. especially since it's a continuous loop.

wbeltran777: Gotcha, I might end up changing the rotation ID to not a constant and generate a random one every time the pedestals are enabled again. I think the cars stop spinning because the const rotation ID is killed

Assets/_PartyParkade/Scenes/DoozyMainMenu.unity
wbeltran777: The position if the pedestals carry over to the scenes so I moved the pedestals to x,y,z=-200

Assets/_PartyParkade/Script/UI/MainMenu/PlayerSelection/3DModelView/ModelCollectionViewController.cs
m4rtint: Also I have an idea on why you are doing this but wondering about your thought process

How come youre making a private static instance?

Assets/_PartyParkade/Script/UI/MainMenu/PlayerSelection/3DModelView/ModelCollectionViewController.cs
```diff
@@ -20,9 +20,21 @@ public class ModelCollectionViewController : MonoBehaviour
...
+            {
+                Destroy(instance.gameObject);
+                instance = this;
+            }
+            else
+            {
+                instance = this;
+            }
+            DontDestroyOnLoad(instance.gameObject);
             SetupMeshViewcontrollers();
```
m4rtint, line 38: can u fix my casing?'

ViewController

Assets/_PartyParkade/Script/UI/MainMenu/PlayerSelection/3DModelView/CarSelectionMeshViewController.cs
m4rtint: looking thru this, that should be a good idea

Assets/_PartyParkade/Script/UI/MainMenu/PlayerSelection/3DModelView/ModelCollectionViewController.cs
wbeltran777: So once I load back on main menu we have:
- A: The pedestals we kept for the end game panel
- B: The pedestals that are always on the menu by default. 

Since we have a DontDestroyOnLoad, then we get both pedestals at once, so we should destroy one. 
The B pedestals are grayed out by default. The A pedestals keep the properties from the previous round. So I choose to destroy A. Destroy(instance.gameobject) 

It's private because there's no need to get the instance from anywhere else.
And it's static so I'm able to check 'this' against the instance.

m4rtint: Nice

mikeslgoh:
[Update the end game panel to match Figma](https://trello.com/c/OCFaZgcO/489-update-the-end-game-panel-to-match-figma)

m4rtint:
Merge Conflicts, Stick with whatever is in develop if you have a choice.

wbeltran777:
Update:
- Moved RotateThisObject to Initialize instead of OnEnable. This makes it easier to make sure we have a different rotation ID for each set of cars. 
- Now we have a RotationID that alternates between 2341 and 2342. When 2341 gets killed, it won't affect the newly created car pedestals using RotationID 2342 (and vice versa).

Assets/_PartyParkade/Script/UI/MainMenu/PlayerSelection/3DModelView/CarSelectionMeshViewController.cs
```diff
@@ -16,7 +16,8 @@ namespace _PartyParkade.Script.UI.MainMenu.PlayerSelection._3DModelView
     /// </summary>
     public class CarSelectionMeshViewController : MonoBehaviour
     {
-        private const int RotationId = 2341;
+        private static int RotationId = 2341;
```
Gursaanj, line 19: should change to m_rotationID?

Assets/_PartyParkade/Script/UI/MainMenu/PlayerSelection/3DModelView/CarSelectionMeshViewController.cs
```diff
@@ -56,6 +58,16 @@ public void UpdateCarModel(int pointer)
...
         }
 
+        /// <summary>
+        /// Called when cars are initialized. Everytime 3Dviewer is created we 
+        /// hop between either RotationID 2341 and 2342
+        /// </summary>
+        public void UpdateRotationID()
+        {
+            RotationId += m_IdAlternator;
+            m_IdAlternator = (-1) * m_IdAlternator;
```
Gursaanj, line 68: dont think -1 needs parenthesis

Assets/_PartyParkade/Script/UI/MainMenu/PlayerSelection/3DModelView/ModelCollectionViewController.cs
```diff
@@ -44,6 +53,8 @@ private void SetupMeshViewcontrollers()
         /// <param name="hats"></param>
         public void Initialize(CarSelections selections, CarHats hats)
         {
+            if(m_meshControllers.Length > 0) m_meshControllers[0].UpdateRotationID();
```
Gursaanj, line 56: add CB after if statement

also would there be any need to null check m_meshControllers[0]?

wbeltran777: Sorry what do you mean by CB?

wbeltran777: You mean space bar yeah? If so fixed

Also the array gets populated on Awake and Initialize is called on Start so I'm not sure we super need to null check after checking the length.

Gursaanj: like if(m_meshControllers.Length > 0)
{
     m_meshControllers[0].UpdateRotationID;
}

Sounds good about the null check

wbeltran777: Oh gotcha gotcha! Will do

Assets/_PartyParkade/Script/UI/MainMenu/PlayerSelection/3DModelView/CarSelectionMeshViewController.cs
```diff
@@ -56,6 +58,16 @@ public void UpdateCarModel(int pointer)
             }
         }
 
+        /// <summary>
+        /// Called when cars are initialized. Everytime 3Dviewer is created we 
+        /// hop between either RotationID 2341 and 2342
+        /// </summary>
+        public void UpdateRotationID()
```
m4rtint, line 65: no need to make this pubic,  just call this in the initialize

Assets/_PartyParkade/Script/UI/MainMenu/PlayerSelection/3DModelView/CarSelectionMeshViewController.cs
```diff
@@ -56,6 +58,16 @@ public void UpdateCarModel(int pointer)
...
+        {
+            m_rotationID += m_IdAlternator;
+            m_IdAlternator = -1 * m_IdAlternator;
```
m4rtint, line 68: If you want to alternate, you can just do
m_rotationID = -m_rotationID

No need to make an extra variable

Assets/_PartyParkade/Script/UI/MainMenu/PlayerSelection/3DModelView/ModelCollectionViewController.cs
```diff
@@ -44,6 +53,11 @@ private void SetupMeshViewcontrollers()
         /// <param name="hats"></param>
         public void Initialize(CarSelections selections, CarHats hats)
         {
+            if (m_meshControllers.Length > 0) 
+            {
+                m_meshControllers[0].UpdateRotationID();
```
m4rtint, line 58: Maybe put this under "cleanup", not sure if that's better.

and make UpdateRotationID static. 

So instead of grabbing [0], you can just grab CarSelectionmeshViewController

Assets/_PartyParkade/Script/UI/MainMenu/PlayerSelection/3DModelView/CarSelectionMeshViewController.cs
wbeltran777: Wasn't sure if a negative id was valid, I'll try that out

Assets/_PartyParkade/Script/UI/MainMenu/PlayerSelection/3DModelView/ModelCollectionViewController.cs
wbeltran777: Will do, I was trying to avoid having any public static methods but I guess this one wouldn't be too bad

PR #185, opened 2020-05-15
Feature/Input Choosing Tool

DJSteins:
Was meant to be used for debugging honking, but now can be used to select the input for each player from within a scene. Just make sure to add the InControl prefab to the scene, and add the InputDebugger script to a game object in the scene.


DJSteins:
[Input Manager Debugger](https://trello.com/c/5xF3UgZ1/549-input-manager-debugger)

DJSteins:
We may also be able to adapt the PlayerController.ChangeInput method so that if a user unplugs/plugs in a device during a game they can switch their input.

Assets/_PartyParkade/Script/Game/Car/CarControllerBehaviour.cs
```diff
@@ -9,6 +9,7 @@
 using _PartyParkade.Script.ScriptableObject.Hats;
 using _PartyParkade.Script.ScriptableObject.Stats;
 using _PartyParkade.Script.UI.MainMenu.PlayerSelection.PlayerSelectionVC.CarAppearance;
+using InControl;
```
m4rtint, line 12: do you still need this?

Assets/_PartyParkade/Script/Game/Inventory.cs
```diff
@@ -23,6 +23,12 @@ public class Inventory : MonoBehaviour
         /// </summary>
         public event Action<PowerUpProperties> OnEquip;
 
+        public IInput InputManager
```
m4rtint, line 26: Why do you need this in the inventory?

m4rtint: Same with Car Controller Behaviour

DJSteins: Yeah I use it in PlayerController.ChangeInput to change the input manager.

DJSteins: Because when you change the input manager for player controller you need to actually propagate that change to everything that has an input manager attached.

DJSteins: Unless I modify it so we update the PlayerControllerInformation inside of InputManager instead of creating a new InputManager in ChangeInput

Assets/_PartyParkade/Script/Data/Tools/InputDebugger.cs
```diff
@@ -0,0 +1,100 @@
...
+    public Choices Choice { set { m_choice = value; } }
+
+    /// <summary>
+    /// God mode inputs. Always starts with keys, but does not give the player keys
+    /// </summary>
+    /// <param name="player">the player who this input choices belongs to.</param>
+    public InputChoice(Player player)
+    {
+        m_playerNumber = (int)player;
+        m_choice = Choices.KEYS;
```
wbeltran777, line 53: Hopped on the branch and it works great! 
Only thing is that on starting a scene the inputs shown on the DEBUG don't reflect the default inputs on each car. Might be worth setting the m_choice to the player's input if that's not too cumbersome

DJSteins: Done!

PR #186, opened 2020-05-18
Feature/end game panel adding ui

wbeltran777:
- Replaced the old car background for the newer one Jocelyn made
- Modified the car panel prefab, removing the old panel, assigning the car images and changing some fonts. 
- Definitely needed to modify the Generic UI scene since I had to move the panels around.


wbeltran777:
> Looks good, whats the change with GenericGameUI.Unity scene?

Though the car panels are prefabed, one of them needs to be bigger as the design requires. So I gotta move one of the objects, which is a scene change and not a prefab change

m4rtint:
[End Game Panel: Display cars](https://trello.com/c/OCFaZgcO/489-end-game-panel-display-cars)

mikeslgoh:
[End Game Panel: Match design to the Figma mocks](https://trello.com/c/I9EUcjEd/548-end-game-panel-match-design-to-the-figma-mocks)

PR #189, opened 2020-05-19
Quick fix for Input Debugger

m4rtint:
I would get errors that Dictionary Key is not found.


Assets/_PartyParkade/Script/Data/Tools/InputDebugger.cs
```diff
@@ -51,7 +51,15 @@ public class InputChoice
     public InputChoice(Player player)
     {
         m_playerNumber = (int)player;
-        PlayerControllerInformation info = GameData.PlayerInformationList[player].PlayerControllerInformation;
+
+        Dictionary<Player, PlayerInformation> playerInformationList = GameData.PlayerInformationList;
+        playerInformationList.TryGetValue(player, out var playerInformation);
```
Gursaanj, line 56: I forget, are we okay with using "var" now? Remembering talking about it but dont remember if we settled on it or not?

m4rtint: forgot to change it back

mikeslgoh:
[No hat and crown](https://trello.com/c/x5n9dXjK/571-no-hat-and-crown)

PR #190, opened 2020-05-19
Installing the Debug Console from the store

m4rtint:
It's not placed into the scene yet. 
this is just the download

Just approve or comment if you want this in or not


m4rtint:
@mikeslgoh yes. so we get to see any errors popped up on builds. 
I haven't integrated it yet, just downloaded

m4rtint:
@Gursaanj too much in terms of space or what?

m4rtint:
@Gursaanj you should try out this branch first. and then see if it's too much

mikeslgoh:
> Having the errors being shown in the builds would be cool, but im still unsure if its too much or not!! Will probably have to test it out fully before comitting to it?!?

I think it could be helpful since whenever we do builds we never know what's happening. How easy is the setup after downloading the plugin?

m4rtint:
> > Having the errors being shown in the builds would be cool, but im still unsure if its too much or not!! Will probably have to test it out fully before comitting to it?!?
> 
> I think it could be helpful since whenever we do builds we never know what's happening. How easy is the setup after downloading the plugin?

It's just a prefab u drag and drop

PR #191, opened 2020-05-20
Feature/switch player effect

DJSteins:
- Adds player launching for switch players with particle effects.
- Fixes cars getting stuck on their sides.
- Fixes AudioManager length check bug


Assets/_PartyParkade/Prefabs/Player/Car/PlayerCar.prefab
```diff
@@ -338,7 +338,7 @@ Rigidbody:
   m_PrefabAsset: {fileID: 0}
   m_GameObject: {fileID: 521212095979244328}
   serializedVersion: 2
-  m_Mass: 10000
+  m_Mass: 100
```
m4rtint, line 341: This is a very big change. does this affect driving?

DJSteins: No driving doesn't use the mass. Actually nothing really does tbh.

Assets/_PartyParkade/Script/Game/Car/CarBoost.cs
```diff
@@ -68,7 +68,7 @@ public void ActivateBoost()
         /// Updates the boost emission
         /// </summary>
         /// <param name="emit">true if should emit, false otherwise</param>
-        private void EmitBoost(bool emit)
+        public void EmitBoost(bool emit)
```
m4rtint, line 71: How come this is public? don't seem to see it used anywhere else

DJSteins: oops lemme put that back to private

Assets/_PartyParkade/Script/Game/Car/CarController.cs
```diff
@@ -146,21 +146,49 @@ public Vector3 GetVelocityChange(Vector3 velocity, float verticalMovement, Vecto
...
+        {
+            LerpTo(Vector3.up);
+        }
+
         /// <summary>
         /// Calculates the direction that the car should tilt to so that it 
         /// remains normal to the surface it is driving onto.
         /// </summary>
-        public Quaternion GetRotationToGround(Transform carRigidBodyTransform)
+        public Quaternion GetRotationToGround(Transform carRigidBodyTransform, Vector3 size)
```
Gursaanj, line 170: public methods above private one

Assets/_PartyParkade/Script/Game/Car/CarController.cs
```diff
@@ -146,21 +146,49 @@ public Vector3 GetVelocityChange(Vector3 velocity, float verticalMovement, Vecto
...
+            {
+                if (groundRay.distance > size.y)
+                {
+                    LerpToUp();
+                }
+                else
+                {
+                    // rotate to a tilted ground
+                    LerpTo(groundRay.normal);
+                }
```
Gursaanj, line 188: I think you shorten the if statement to this?

`if (!Physics.Raycast(carRgidBody Transform.position, -m_curNormal, out var groundRay) || groundRay.distance > size.y)
{
         LerpToUp();
}
else
{
          LerpTo(groundRay.normal;);
}`

Assets/_PartyParkade/Script/Managers/AudioManager.cs
```diff
@@ -61,6 +61,11 @@ public void PlayOneShotSFX(SOUNDEFFECT soundEffect, float relativeVolume = 1.0f)
 
         public void PlayOneShotPowerUpUsage(AudioClip[] soundEffects, float relativeVolume = 1.0f)
         {
+            if (soundEffects.Length == 0)
+            {
+                Debug.Log("Sound effects array in PlayOneShotPowerUpUsage is empty.");
+                return;
+            }
```
Gursaanj, line 68: Ill try to find and place a temporary SFX for the powerup  in the coming days until we decide on something more permanent

Assets/_PartyParkade/Script/PowerUp/SwitchAllPlayers.cs
```diff
@@ -7,19 +7,185 @@
...
+
+    private List<TrackingInfo> m_info;
+
+    /// <summary>
+    /// Use this class to track 
+    /// </summary>
+    private class TrackingInfo
+    {
+        // the car
+        public CarControllerBehaviour car;
```
m4rtint, line 38: formatting

Assets/_PartyParkade/Script/PowerUp/SwitchAllPlayers.cs
```diff
@@ -7,19 +7,185 @@
...
+        }
+    }
+
+    private void OnEnable()
+    {
+        BaseWorldManager.OnFixedUpdate += UpdateLaunchedCars;
+        m_info = new List<TrackingInfo>();
+    }
+
+    private void OnDestroy()
```
m4rtint, line 92: OnDisable is better.

Assets/_PartyParkade/Script/PowerUp/SwitchAllPlayers.cs
```diff
@@ -7,19 +7,185 @@
...
+    }
+
+    /// <summary>
+    /// Plays then destroys an instantiated particle system
+    /// </summary>
+    /// <param name="particle">the particle system to instantiate</param>
+    /// <param name="position">the position to activate at</param>
+    private void PlayOnce(ParticleSystem particle, Vector3 position)
+    {
+        ParticleSystem obj = Instantiate(particle, position, Quaternion.identity);
```
m4rtint, line 155: It's better to play the Particle System on the Vehicle instead of Instantiating each time.

m4rtint:
Can you unpack all the particle effects prefab and save them as its own in our own Prefab/particleeffects folder. Instead of directly editting the EpicToonFX prefabs given.

DJSteins:
@m4rtint I thought I did unpack them? I specifically went and clicked completely unpack on each toon fx I used

Assets/_PartyParkade/Script/PowerUp/SwitchAllPlayers.cs
DJSteins: But if they are attached to the vehicle then they move with the vehicle as the vehicle moves. Or will they not if I attach to the player prefab, at the same level as the rigid body?

Assets/_PartyParkade/Script/Game/Car/CarController.cs
DJSteins: I think I tried that already but I can try again

m4rtint:
> @m4rtint I thought I did unpack them? I specifically went and clicked completely unpack on each toon fx I used

Hmm cuz on Github it shows that you edited the original file, and there's no new particle effects files added.

Assets/_PartyParkade/Script/PowerUp/SwitchAllPlayers.cs
m4rtint: There's a way to change the particle effect to be in World position instead of local position (which follows the parent)

m4rtint: BUT. because we aren't instantiating that much that fast. It shouldn't be a problem (this is more of a memory management thing. I honestly would be totally fine leaving it as is

@Gursaanj thoughts on this practice.

Gursaanj: hmmm good point, I think it should be fine leaving it as is, but if later on someone wants to make it more memory efficient then that should be fine too!!

mikeslgoh:
[(2) Switch Player Powerup: Transition fade in/out](https://trello.com/c/ZtHRVXUG/566-2-switch-player-powerup-transition-fade-in-out)

Assets/_PartyParkade/Script/PowerUp/SwitchAllPlayers.cs
DJSteins: what's wrong with the formatting?

m4rtint: oh nvm it's a publish variable

Assets/_PartyParkade/Script/PowerUp/SwitchAllPlayers.cs
m4rtint: OnEnable could be called multiple times before the OnDestroy is called.

DJSteins: disable gets called before destroy. If I switch to using OnDisable, then the update function is never called.

DJSteins: I'm switching onEnable to Awake

m4rtint: how come the update function is never called if using OnDisable?

DJSteins: My guess is the order of operations goes OnEnable -> OnDisable -> OnDestroy and maybe OnDisable gets called as soon as the power-up is activated? So if I remove the update function in OnDisable, then the update is never called.

m4rtint: okay that makes sense. 
yeah Then putting it on Awake is the better idea.

PR #198, opened 2020-05-22
Options Panel - Pre-Fab and Scene updated

jclyn-t:
Successfully revised and anchored option panel pre-fab, replaced old options panel scene with new pre-fab


mikeslgoh:
[(1) Update option panel based on Figma](https://trello.com/c/EKMSPv5Z/494-1-update-option-panel-based-on-figma)

m4rtint:
image

PR #205, opened 2020-05-30
Pick Up Item When On Top On Spawn

m4rtint:
Put whatever is on OnTriggerEnter is also in OnTriggerStay


m4rtint:
[Can't get power up as it spawns](https://trello.com/c/5aR1oidP/589-cant-get-power-up-as-it-spawns)

PR #207, opened 2020-05-31
Quick Bug Fixes

m4rtint:
- Floor height decrease so players don't float
- Gravity Well set to 7 seconds
- Prettify the sea wall


mikeslgoh:
[No Name Island: Players floating above ground](https://trello.com/c/w032sixH/594-no-name-island-players-floating-above-ground)

mikeslgoh:
[Gravity Well does not disappear](https://trello.com/c/u4cQoqrp/595-gravity-well-does-not-disappear)

PR #208, opened 2020-05-31
Generic UI Debugger

m4rtint:
Script to add generic UI when playing from map scene if needed/enabled


Assets/_PartyParkade/Script/Data/Tools/GenericUIDebugger.cs
```diff
@@ -0,0 +1,32 @@
...
+        [SerializeField] private bool m_loadInGenericUI = true;
+        private void Start()
+        {
+            if (m_loadInGenericUI)
+            {
+                int countLoaded = SceneManager.sceneCount;
+                if (countLoaded <= 1)
+                {
+                    SceneManager.LoadScene(1, LoadSceneMode.Additive);
+                }
```
DJSteins, line 28: What does `1` represent in `LoadScene(1, LoadSceneMode.Additive)`?

m4rtint: The generic UI scene. I'll put it on top

DJSteins: Do you think we should be representing scene numbers as an enum somewhere?

Gursaanj: So when we build our game, we first must add the scenes into out build settings 

image


so 1 refers to the index representation of The GenericGameUI Scene according to the Build and the SceneManager

DJSteins: ohhh okay I see. Didn't know about that.

m4rtint: @DJSteins We used to have the index stored in a static class. We could bring it back for important scenes. 

Since the MainMenu and the GenericGameUI, should technically always be the same places. (Or at least it shouldn't be moving). 

If we bring back that enum ,it should be another refactor task

PR #211, opened 2020-06-01
Options Panel - Slider Handle and Titles updated

jclyn-t:
Slider handle bug resolved (will not stretch into oval/stays circular). Titles also moved slightly to the left to visually make room for the handles.


mikeslgoh:
Comments from Martin:
I strongly recommend, deleting this branch and checking what we already have.
Then edit according to what she thinks is neccassary.

PR #212, opened 2020-06-01
Feature/scriptable fx

DJSteins:
Makes scriptable objects for sound effects that are then attached to the audio manager


Assets/_PartyParkade/Script/Data/SoundEffects.cs
```diff
@@ -64,6 +19,7 @@ public enum AUDIOGROUPS
         BGM,
         SwitchPlayers,
         OilSpill,
-        GravityWell
+        GravityWell,
+        Hjonk
```
m4rtint, line 23: ?lol

PR #214, opened 2020-06-02
Updated Hats

m4rtint:


m4rtint:
[No crown and wizard hat icon](https://trello.com/c/2YQ3G2na/574-no-crown-and-wizard-hat-icon)

PR #219, opened 2020-06-03
Fix/bg music

DJSteins:
Trying to fix corrupt audio files


mikeslgoh:
[(0.5) Refine background music](https://trello.com/c/nAZj68ID/606-05-refine-background-music)

PR #221, opened 2020-06-04
Bugfix/powerup reset on replay

mikeslgoh:
- reset player rotation on restart
- powerups do not carry over when game is restarted (either from end game or pause panel)
- car hats are back in the player selection


mikeslgoh:
[Powerup follow over to next game if click replay](https://trello.com/c/h5EwY1Ue/597-powerup-follow-over-to-next-game-if-click-replay)

Assets/_PartyParkade/Script/Game/PlayerController.cs
```diff
@@ -57,6 +57,15 @@ public void ResetPlayer(Vector3 spawnPoint)
         {
             Car.SetCarPosition(spawnPoint);
             SetPointText(0);
+            ResetInventory();
+        }
+
+        private void ResetInventory()
+        {
+            if (m_inventory)
```
m4rtint, line 65: does this work on C#? it's doing a null check right?

mikeslgoh: Yeah I'm doing a null check. Not sure if this should work for C#. Should I do != null to be safe?

m4rtint: You can do some research on this, i'm trusting the compiler, but u should do some research on this.

mikeslgoh: https://forum.unity.com/threads/optimizing-null-check-null-vs-bool-out-functions.482118/

Near the end of the thread, there is apparently a method called ReferenceEquals(UnityObject, null) that might be better in terms of performance as well

m4rtint: interesting, that's good to know. 
But this isn't called all the time, but we know this work right?

mikeslgoh: yup it works

Gursaanj: This definitely works!

Gursaanj: Most people just do the != null for readability so no one confuses the variable for a bool, from what I understand

m4rtint: ah that's fine, but good to know C# accepts it.

PR #223, opened 2020-06-04
Enhancement/debug parent class

Gursaanj:
Debugger Windwo

Updated Debugger Window for builds : Now allows user to decide which Debugger to open from from the main panel 

Created DebugLogic Parent class : To create the functionality used by all debugging logic for builds. Also creates a back button to go back to the main "page" of the Debugger Window

Created a PowerupDebugger Logic for users to change their Powerups manually in build


Gursaanj:
[(3) Update the powerup tool to be used as a build debugger](https://trello.com/c/EHET12kp/567-3-update-the-powerup-tool-to-be-used-as-a-build-debugger)

Gursaanj:
[(2) Parent class for build debugger tools](https://trello.com/c/PxGANzLM/591-2-parent-class-for-build-debugger-tools)

Assets/_PartyParkade/Script/Debug/Debug Logic/ParkingDebuggerLogic.cs
```diff
@@ -291,22 +257,13 @@ private void FocusOnParkingSpot(GameObject parkingSpot)
...
-        m_playerSelectionRect = CreateRect(m_playerSelectionRect, m_masterRect.x, m_titleRect.height + m_scrollViewRect.height, m_masterRect.width, 2 * heightUnit);
-        m_lowerRect = CreateRect(m_lowerRect, m_masterRect.x, m_titleRect.height + m_scrollViewRect.height + m_playerSelectionRect.height, m_masterRect.width, m_masterRect.height - m_titleRect.height - m_scrollViewRect.height - m_playerSelectionRect.height);
+        m_titleRect = CreateRect(m_titleRect, m_baseRect.x, m_baseRect.y, m_baseRect.width, 3.3f * heightUnit);
+        m_scrollViewRect = CreateRect(m_scrollViewRect, m_baseRect.x, m_titleRect.height, m_baseRect.width, 5.7f * heightUnit);
+        m_playerSelectionRect = CreateRect(m_playerSelectionRect, m_baseRect.x, m_titleRect.height + m_scrollViewRect.height, m_baseRect.width, 2 * heightUnit);
+        m_lowerRect = CreateRect(m_lowerRect, m_baseRect.x, m_titleRect.height + m_scrollViewRect.height + m_playerSelectionRect.height, m_baseRect.width, m_baseRect.height - m_titleRect.height - m_scrollViewRect.height - m_playerSelectionRect.height);
 
         m_subHeaderHeight = 0.5f * heightUnit;
     }
```
DJSteins, line 268: What are all of the magic numbers in here? 12, 3.3f, 5.7f, 0.5f ... They should be assigned to const variables at the very least.

m4rtint: If you want to make it more organized. You can create a private struct with all The constants inside within this class

Assets/_PartyParkade/Script/Debug/Debug Logic/PowerupDebuggerLogic.cs
```diff
@@ -0,0 +1,145 @@
...
+    private float[] m_sliderValues;
+    private int m_numberOfPowerups;
+
+    public PowerupDebuggerLogic(Rect masterRect, bool isDebugBuild) : base(masterRect, isDebugBuild) { }
+
+    public override void InitializeData()
+    {
+        m_players = PlayerManager.Players != null ? PlayerManager.Players : null;
+        m_sliderValues = new float[m_players.Count];
+        m_numberOfPowerups = Enum.GetNames(typeof(POWERUPS)).Length;
```
DJSteins, line 30: Why not use a List<float> for `m_sliderValues`?

Assets/_PartyParkade/Script/Debug/Debug Logic/PowerupDebuggerLogic.cs
```diff
@@ -0,0 +1,145 @@
...
+    private List<PlayerController> m_players;
+    private float[] m_sliderValues;
+    private int m_numberOfPowerups;
+
+    public PowerupDebuggerLogic(Rect masterRect, bool isDebugBuild) : base(masterRect, isDebugBuild) { }
+
+    public override void InitializeData()
+    {
+        m_players = PlayerManager.Players != null ? PlayerManager.Players : null;
+        m_sliderValues = new float[m_players.Count];
```
DJSteins, line 29: Why would PlayerManager.Players be null? It's a list isn't it?

Assets/_PartyParkade/Script/Debug/Debug Logic/PowerupDebuggerLogic.cs
Gursaanj: since i wont be adding or removing more elements into the list, might as well as keep it as an array, since I know the length will never change

Assets/_PartyParkade/Script/Debug/Debug Logic/PowerupDebuggerLogic.cs
```diff
@@ -0,0 +1,145 @@
...
+        m_numberOfPowerups = Enum.GetNames(typeof(POWERUPS)).Length;
+
+        for (int i = 0; i < m_sliderValues.Length; i++)
+        {
+            PlayerController player = m_players[i];
+            if (player != null && player.Inventory != null && player.Inventory.PlayerPowerUp != null && player.Inventory.PlayerPowerUp.Properties != null)
```
DJSteins, line 35: Can you abstract this into a method like `bool playerHasPowerUp(PlayerController player)`

Assets/_PartyParkade/Script/Debug/Debug Logic/PowerupDebuggerLogic.cs
Gursaanj: A list can be null!!

Assets/_PartyParkade/Script/Debug/Debug Logic/PowerupDebuggerLogic.cs
```diff
@@ -0,0 +1,145 @@
...
+            {
+                m_sliderValues[i] = (int) m_players[i].Inventory.PlayerPowerUp.Properties.PowerUpEnum;
```
DJSteins, line 37: Can you make this into a method like `int getPowerUpEnum()` that belongs to `PlayerController`?

Assets/_PartyParkade/Script/Debug/Debug Logic/PowerupDebuggerLogic.cs
DJSteins: Yes, but if it's null then we have a bigger problem don't you think? Like that should raise an error in the PlayerManager if its list of Players is null.

Assets/_PartyParkade/Script/Debug/Debug Logic/ParkingDebuggerLogic.cs
Gursaanj: Ill make them into consts, no need for a struct I think

Assets/_PartyParkade/Script/Debug/Debug Logic/PowerupDebuggerLogic.cs
DJSteins: And then we don't have to be checking if its null all over our code because we can be sure that whoever designed the PlayerManager designed it in such a way that errors with its own variables are handled in the PlayerManager class itself.

DJSteins: And if we are doing null checks anywhere in our code like this then we should probably be logging something at the very least to say that its null.

Assets/_PartyParkade/Script/Debug/Debug Logic/PowerupDebuggerLogic.cs
DJSteins: Fair enough. I also see that you assign m_players to null if PlayerManager.Players is null. So getting m_players.Count would throw an error if m_players is null. You should rather make m_players an empty array instead of assigning it to null.

Assets/_PartyParkade/Script/Debug/Debug Logic/PowerupDebuggerLogic.cs
```diff
@@ -0,0 +1,145 @@
...
+        CreateRemoveAllButton();
+    }
+
+    private void DrawRects()
+    {
+        float heightUnit = m_baseRect.height / 6;
+
+        m_titleRect = CreateRect(m_titleRect, m_baseRect.x, m_baseRect.y, m_baseRect.width, heightUnit);
+        m_selectionRect = CreateRect(m_selectionRect, m_baseRect.x, m_titleRect.height, m_baseRect.width, 4 * heightUnit);
+        m_removeAllRect = CreateRect(m_removeAllRect, m_baseRect.x, m_titleRect.height + m_selectionRect.height, m_baseRect.width, heightUnit);
```
DJSteins, line 62: Magic number 4 and 6. Please assign to variables or add comments.

Assets/_PartyParkade/Script/Debug/Debug Logic/PowerupDebuggerLogic.cs
```diff
@@ -0,0 +1,145 @@
...
+        GUILayout.BeginArea(m_titleRect);
+        GUILayout.Label("PowerUp Debugger", m_skin.GetStyle(m_titleSkin));
+        GUILayout.EndArea();
+    }
+
+    private void CreatePowerUpSelection()
+    {
+        if (m_players == null)
+        {
+            Debug.Log("No Players Present");
```
DJSteins, line 76: I think this is an Error level log. Our game should never have zero players.

Assets/_PartyParkade/Script/Debug/Debug Logic/PowerupDebuggerLogic.cs
```diff
@@ -0,0 +1,145 @@
...
+            return;
+        }
+
+        GUILayout.BeginArea(m_selectionRect);
+
+        GUILayout.BeginVertical();
+
+        for (int i = 0, len = m_players.Count; i < len; i++)
+        {
+            if (m_players[i].Inventory == null)
```
DJSteins, line 86: m_players[i] is reused a lot. You should assign it to a variable so it isn't reindexed over and over.

Assets/_PartyParkade/Script/Debug/Debug Logic/PowerupDebuggerLogic.cs
```diff
@@ -0,0 +1,145 @@
...
+                continue;
+            }
+
+            GUILayout.BeginHorizontal();
+
+            GUILayout.Label(m_players[i].name, m_skin.GetStyle(m_subHeaderSkin));
+
+            GUILayout.BeginVertical();
+
+            GUILayout.Label(Enum.IsDefined(typeof(POWERUPS), (int) m_sliderValues[i]) ? Enum.GetName(typeof(POWERUPS), (int) m_sliderValues[i]) : "None" , m_skin.GetStyle(m_subHeaderSkin));
```
DJSteins, line 97: Line too long. Break into multiple lines.

Assets/_PartyParkade/Script/Debug/Debug Logic/PowerupDebuggerLogic.cs
```diff
@@ -0,0 +1,145 @@
...
+
+            m_sliderValues[i] = Mathf.RoundToInt(GUILayout.HorizontalSlider(m_sliderValues[i], 0, m_numberOfPowerups));
+
+            GUILayout.EndVertical();
+
+            if (GUILayout.Button("Give Powerup", m_skin.GetStyle(m_subHeaderButtonSkin)))
+            {
+                if (PowerUpsObjectPooler.Instance != null && Enum.IsDefined(typeof(POWERUPS), (int)m_sliderValues[i]))
+                {
+                    PowerUp powerUp = PowerUpsObjectPooler.Instance.SpawnFromPool((POWERUPS)(int) m_sliderValues[i]);
```
DJSteins, line 107: Why is m_sliderValues an array of floats if you are always casting to int?

Assets/_PartyParkade/Script/Debug/Debug Logic/PowerupDebuggerLogic.cs
```diff
@@ -0,0 +1,145 @@
...
+        GUILayout.BeginArea(m_removeAllRect);
+        GUILayout.FlexibleSpace();
+
+        if (GUILayout.Button("Reset All Powerups?", m_skin.GetStyle(m_buttonSkin)))
+        {
+            for (int i = 0, len = m_players.Count; i < len; i++)
+            {
+                PlayerController player = m_players[i];
+
+                if (player != null && player.Inventory != null)
```
DJSteins, line 135: A player probably shouldn't be stored in m_players if it is null. Then you don't need to do `player != null` here.

Assets/_PartyParkade/Script/Debug/DebugWindow.cs
```diff
@@ -39,30 +59,45 @@ private void Update()
...
+        }
     }
 
     /// <summary>
     /// Initialize All Relevant data pertaining to the debugging Behaviour
     /// </summary>
     private void InitializeData()
     {
-        m_logic.InitializeData();
+        if (m_debuggingLogic != null)
```
DJSteins, line 80: Any reason this would be null if you are already initializing it to a new list in Awake?

Assets/_PartyParkade/Script/Debug/DebugWindow.cs
```diff
@@ -39,30 +59,45 @@ private void Update()
...
+        }
     }
 
     /// <summary>
     /// Lose all relevant data pertaining to the debugging behaviour
     /// </summary>
     private void ResetData()
     {
-        m_logic.ResetData();
+        if (m_debuggingLogic != null)
```
DJSteins, line 91: Same here. Any reason this would be null?

Assets/_PartyParkade/Script/Debug/DebugWindow.cs
```diff
@@ -74,6 +109,47 @@ private void ClosePanel()
...
+        GUILayout.BeginHorizontal();
+        GUILayout.FlexibleSpace();
+        GUILayout.BeginVertical();
+
+        foreach (Type type in m_listOfDebuggingOptions)
+        {
+            if (GUILayout.Button(type.Name))
+            {
+                ConstructorInfo constructorInfo = null;
+                object[] parameters = null;
```
DJSteins, line 125: I think it's better practice here to assign each parameter to its own variable rather than storing them in an array of arbitrary objects.

Assets/_PartyParkade/Script/Debug/Debug Logic/PowerupDebuggerLogic.cs
m4rtint: I think that even though you never set it to null, you should still do null checks. 

It's part of defensive programming where, you never expected it to be null, but just incase it is null, it'll still run.

Assets/_PartyParkade/Script/Debug/DebugWindow.cs
DJSteins, line 91: If you are doing null checks for m_debuggingLogic then you should probably add a Debug/Error log line if it is null.

Assets/_PartyParkade/Script/PowerUp/BasePowerUp/IPowerUp.cs
```diff
@@ -11,8 +11,7 @@ public interface IPowerUp
 {
     void Activate(PlayerController player);
     void Deactivate();
-    /* Decide how to return an animation */
-    // void GetAnimation();
+    //AudioClip[] GetSounds();
```
DJSteins, line 14: Just delete GetSounds

Assets/_PartyParkade/Script/Debug/Debug Logic/PowerupDebuggerLogic.cs
DJSteins: We should be logging in cases where it is null then. Like add an `else` statement to do a Debug.Log or Error.Log or something

Assets/_PartyParkade/Script/Debug/Debug Logic/PowerupDebuggerLogic.cs
m4rtint: If it's Error level, will it crash? 

Cuz is it possible to run this script during the Main menu where m_players will be null?

Assets/_PartyParkade/Script/Debug/Debug Logic/PowerupDebuggerLogic.cs
Gursaanj: good catch, will probably just null check before doing the for loop!

Assets/_PartyParkade/Script/Debug/Debug Logic/PowerupDebuggerLogic.cs
Gursaanj: yup, an else to Debu.log sounds like the way to go!!

Assets/_PartyParkade/Script/Debug/Debug Logic/PowerupDebuggerLogic.cs
Gursaanj: This is more for, if we (the programmers) start the game without the proper flow, this is a defense check!

Assets/_PartyParkade/Script/Debug/Debug Logic/PowerupDebuggerLogic.cs
Gursaanj: Because GUILayout.Horizontal slider can only return float (super frustrating). so the result must be a conversion from float to int (then to enum)

Assets/_PartyParkade/Script/Debug/Debug Logic/PowerupDebuggerLogic.cs
Gursaanj: Again defense check, but sure thing!

Assets/_PartyParkade/Script/Debug/DebugWindow.cs
Gursaanj: this is because ContstructorInfo.invoke() only accepts Object[]  but I was thinking the same thing

Assets/_PartyParkade/Script/PowerUp/BasePowerUp/IPowerUp.cs
Gursaanj: Sure thing, it was creating compile errors when I pulled so wasnt certain if another branch needed it!!!

Assets/_PartyParkade/Script/Debug/Debug Logic/PowerupDebuggerLogic.cs
Gursaanj: Will make it warning level!

Assets/_PartyParkade/Script/Debug/Debug View/PowerupDebuggerView.cs
```diff
@@ -0,0 +1,167 @@
...
+                    player.Inventory.Discard();
+                    m_sliderValues[i] = m_numberOfPowerups;
+                }
+            }
+        }
+
+        GUILayout.EndArea();
+    }
+
+    private bool PlayerHasPowerUp(PlayerController player)
```
m4rtint, line 163: DoesPlayerHasPowerUp 
Rename it

Assets/_PartyParkade/Script/Debug/Debug View/PowerupDebuggerView.cs
```diff
@@ -0,0 +1,167 @@
...
+    private List<PlayerController> m_players;
+    private float[] m_sliderValues;
+    private int m_numberOfPowerups;
+
+    public PowerupDebuggerView(Rect masterRect, bool isDebugBuild) : base(masterRect, isDebugBuild) { }
+
+    public override void InitializeData()
+    {
+        m_numberOfPowerups = Enum.GetNames(typeof(POWERUPS)).Length;
+        m_players = PlayerManager.Players != null ? PlayerManager.Players : null;
```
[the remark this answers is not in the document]
Gursaanj: oops forgot to remove from my first refactor

PR #224, opened 2020-06-04
Bugfix/Null BG Music

DJSteins:
Needed to rename the BackGroundMusic script to BackgroundMusic and then add the script reference back to the scriptable object.


DJSteins:
> Is renaming it just just a qol?

@m4rtint no the file name was different from the class name. Pretty sure they have to be the same.

PR #225, opened 2020-06-05
Feature/score debugger

m4rtint:
image

- I still need to put this into Debug window. But the logic is here 
- The meta files is from Sirenx/Odin


Assets/_PartyParkade/Script/Data/Tools/ScoreDebugger.cs
```diff
@@ -0,0 +1,44 @@
...
+    public class ScoreDebugger
+    {
+        public void InstantWin()
+        {
+            ScoreManager.SetGameEnd();
+        }
+
+        public void IncrementScore(Player player)
+        {
+            List<PlayerController> allPlayers = PlayerManager.Players;
```
Gursaanj, line 15: It seems weird to get PlayerManager.Players each time you press the button 

Would it be possible to save that list Count onEnable in ScoreDebugger and make a get for it in PlayerScoreDebugger to pass it as a param in IncrementScore 

I know that sounds tedious, but PlayerManager.Players should really only need to be called once

Assets/Editor/PartyParkadeScoreDebugger.cs
```diff
@@ -0,0 +1,53 @@
...
+
+    private void OnDisable()
+    {
+        m_scoreDebugger = null;
+        EditorApplication.playModeStateChanged += OnPlayModeExited;
+    }
+
+    private void DrawHeader()
+    {
+        GUILayout.BeginArea(new Rect());
```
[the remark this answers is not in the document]
m4rtint: something like that. i don't think this script is used right now.

Assets/_PartyParkade/Script/Data/Tools/ScoreDebugger.cs
m4rtint: it's not tedious, it makes sense.

instead of onenable (since onenable might be slower than the PlayerManager loading, I'll make another variable where it's a lazy load

mikeslgoh:
[(2) Create a score debugger for insta win](https://trello.com/c/hwiXOsZe/605-2-create-a-score-debugger-for-insta-win)

Assets/_PartyParkade/Script/Data/Tools/ScoreDebugger.cs
Gursaanj: Sounds good!!

PR #229, opened 2020-06-07
Update IngameDebugConsole.prefab

m4rtint:
- Removed event system that overrides InControl


m4rtint:
[Navigate main menu with controller](https://trello.com/c/IQZIzjfr/627-navigate-main-menu-with-controller)

mikeslgoh:
So is this suppose to be a solution for the controller navigation on the menu? Or is Daniel's work going to be the actual solution? Because there are 2 tickets right now for the controller navigation

m4rtint:
@mikeslgoh he also changed the prefab in the same way.

PR #230, opened 2020-06-07
Enhancments/pause panel controller support

DJSteins:
Adds proper controller support to the pause panel.
- Pressing start button opens pause panel
- Buttons are properly highlighted when navigating up/down through the panel using the left stick or d-pad
- Panels for Paused, Reset, and Return to main menu have been refactored into 3 separate game objects
- Pressing the Start button when on the Reset or Return to main panels will close the pause menu all together
- Pressing the back button (B on Xbox 360) when on the Reset or Return to main panels will return the user to the main pause panel
- Pressing the back button the main pause panel will return the player to the game


mikeslgoh:
[Navigate menu using controller](https://trello.com/c/GnWy6FKe/608-navigate-menu-using-controller)

m4rtint:
image

You'll need to remove the "Button OnClick" from the inspector and set it up from the UIView on Doozy, if not the Graph Flow will break

m4rtint:
image


Can you also refactor this part? The PlayStation controller is currently messed up (spams the start button too fast and breakinng the flow).

I think the best solution is for the "BaseWorldManager" to only allow pause. and the pause panel to only allow Resume (seperating the functionality).

mikeslgoh:
[Back from option doesnt work](https://trello.com/c/9JAsbYWk/628-back-from-option-doesnt-work)

mikeslgoh:
[Using X on Playstation controller to press submit on Menus](https://trello.com/c/2DArSBH4/640-using-x-on-playstation-controller-to-press-submit-on-menus)

wbeltran777:
Keyboard movements aren't working for pause menu navigation, not sure if that's something that's also considered in this PR. 
Also when I open the pause menu in game and then close it by clicking 'Resume', the game freezes. 
I went on develop and the bug isn't on there.

m4rtint:
@wbeltran777 keyboard control wasn't part of the road map, but it could/might be added as another task

I'll fix the game freeze problem. but it's not like Unity crashes right? just that you can't move?

wbeltran777:
> @wbeltran777 keyboard control wasn't part of the road map, but it could/might be added as another task
> 
> I'll fix the game freeze problem. but it's not like Unity crashes right? just that you can't move?

Cool cool. Yeah no the music keeps going and you can hear the engine rumbling, so it's probably a state manager problem.

mikeslgoh:
> @wbeltran777 keyboard control wasn't part of the road map, but it could/might be added as another task
> 
> I'll fix the game freeze problem. but it's not like Unity crashes right? just that you can't move?

I'll create a ticket to add keyboard navigation for the pause menu

m4rtint:
@wbeltran777 
I Fixed it, I did an oopsie.

wbeltran777:
> @wbeltran777
> I Fixed it, I did an oopsie
Cool cool yeah the bug's gone

Assets/_PartyParkade/Script/UI/AbstractListener.cs
```diff
@@ -0,0 +1,33 @@
...
+    /// </summary>
+    /// <returns></returns>
+    protected abstract bool KeyPressed();
+
+    /// <summary>
+    /// Do something when an input is detected
+    /// </summary>
+    protected abstract void HandleInput();
+
+    private void Update()
```
wbeltran777, line 26: Should this Update be subscribed to another global update?

m4rtint: This is the pause menu, so this update has to be running while everything else is paused.

m4rtint: so no.

PR #231, opened 2020-06-08
Fixed Oclusion in winter level

wbeltran777:
- Fixed occlusion bug. Basically the road dividers had a super tall collider on them to stop people from going out of bounds. This gets in the way of the ray casting that activates the CullSphere. Just made them a bit shorter


m4rtint:
@wbeltran777 would it be better (cover all cases) to tag those tall walls as a certain layermask? rather than hoping the walls aren't "too tall" in level design? 

and then in the raycast ignore those masks?

wbeltran777:
> @wbeltran777 would it be better (cover all cases) to tag those tall walls as a certain layermask? rather than hoping the walls aren't "too tall" in level design?
> 
> and then in the raycast ignore those masks?

Hmm yeah you're super right, I'll update the PR with those changes

mikeslgoh:
[(1) Ice wall (bottom map) no transparent/zoom feature](https://trello.com/c/UyjzZjT8/586-1-ice-wall-bottom-map-no-transparent-zoom-feature)

Assets/_PartyParkade/Script/Game/CameraCulling.cs
```diff
@@ -41,7 +42,7 @@ private void CameraCullingUpdate()
 
             if (Physics.Raycast(this.transform.position,
                 (m_cullSphere.transform.position - this.transform.position).normalized,
-                out hit))
+                out hit, Mathf.Infinity, layerMask.value))
```
m4rtint, line 45: if i'm understanding this correctly, you're only casting to the player, and ignoring everything else?

wbeltran777: The layermask is set on the prefab as a serialized field. Basically it only checks the 'Obstacle' and 'PlayerModel' layers.

m4rtint: interesting

Assets/_PartyParkade/Prefabs/Player/Player.prefab
```diff
@@ -480,6 +483,11 @@ PrefabInstance:
       propertyPath: m_Name
       value: PlayerCar
       objectReference: {fileID: 0}
+    - target: {fileID: 6576194232417591580, guid: b7d96f7a8618dcb4393b1823982ed7d7,
```
m4rtint, line 486: what did u change here?

Assets/_PartyParkade/Script/Game/CameraCulling.cs
m4rtint: change it to m_layerMask

Assets/_PartyParkade/Prefabs/Player/Player.prefab
wbeltran777: Added the 'PlayerModel' layer to the car model. So the ray cast will stop at the car. If it's not there then, if the floor is part of an object, you'll get occlusion when you shouldn't need it.

PR #232, opened 2020-06-11
Message Boxes

m4rtint:
image


Assets/_PartyParkade/Script/PowerUp/DriveText/MessageBoxBehaviour.cs
```diff
@@ -0,0 +1,61 @@
...
+    /// <summary>
+    /// Sets the Text and Image for a text message
+    /// </summary>
+    /// <param name="text">text for the text box</param>
+    /// <param name="img">image to place in the message</param>
+    [Button]
+    public void Set(string text, Sprite img = null)
+    {
+        TextMessage.text = text;
+        if (img)
```
Gursaanj, line 43: for readability, it will be better to be explicit for the if statement, if (img != null)

mikeslgoh: I did this for my bugfix PR and it works the same?

mikeslgoh:
[(2) Create message bubble that stretches (dynamic object)](https://trello.com/c/ZjVlTONz/617-2-create-message-bubble-that-stretches-dynamic-object)

Assets/_PartyParkade/Script/PowerUp/DriveText/MessageBoxBehaviour.cs
m4rtint: yeah

Gursaanj: For sure it will work the same, but after thinking about it, if we ever have like a lot of code, for the sake of the others, not needing to think twice if a variable is a bool or not seems like the better option

PR #235, opened 2020-06-12
BugFix : Player Selection reverts to default values.

m4rtint:
- Placed Updating Material lower since that's how the flow supposed to work.
- Each car has its own material, so need to update car before update material


m4rtint:
[Player selection goes back to default cars color when player goes back from map selection to player selection screen](https://trello.com/c/l2Ajlk09/602-player-selection-goes-back-to-default-cars-color-when-player-goes-back-from-map-selection-to-player-selection-screen)

PR #238, opened 2020-06-14
Sorting Layer set

m4rtint:
Current Sorting Layer:
- ParticleEffects
- Default
- UserInterface


m4rtint:
[Vortex power-up is on top of the "Power-up box" ui](https://trello.com/c/YtTNLADK/599-vortex-power-up-is-on-top-of-the-power-up-box-ui)

PR #240, opened 2020-06-15
Parking spots were missing particle effect

mikeslgoh:
Some of the spots had the particle effects field empty which was causing the bug where all the spots lit up


mikeslgoh:
[(0.5) Parking Spot Delay Disable](https://trello.com/c/O3xMNhUl/644-05-parking-spot-delay-disable)

mikeslgoh:
> So I take it that this bug was not appearing on No Name Island??

Yeah was walking through the build with Kevin earlier and he noticed that only No Name island had no issues.

m4rtint:
> > So I take it that this bug was not appearing on No Name Island??
> 
> Yeah was walking through the build with Kevin earlier and he noticed that only No Name island had no issues. 

Hmm a good experiment is to Change the particle effect through the prefab only.. And see how that effects
All the objects in the map. I supposedly referenced the particle fx in the prefab and theoretically it should apply to all of them in all maps.

PR #242, opened 2020-06-15
Text and Drive prefab and deployable

wbeltran777:
- Created Text and Drive scriptable object
- Added new method in Deployable for interrupting timer. This is useful for writing stuff that you pick up. 
- Place holder sphere for the text and drive, and current text and drive icon is just the crown icon. 
- Updated power up pooler with text and drive prefab


Assets/_PartyParkade/Script/PowerUp/BasePowerUp/Deployable.cs
```diff
@@ -147,6 +147,14 @@ private void OnTriggerEnter(Collider other)
         }
     }
 
+    /// <summary>
+    /// Interrupt power up duration
+    /// </summary>
+    protected void InterruptTimer()
```
m4rtint, line 153: I always take it interrupt means stop. 

is this resetting the timer?

wbeltran777: True I was going back and forth on naming conventions. 
This sets the timer to it's max amount so the item deactivates. 
Maybe CancelTimer would make more sense.

m4rtint: so looking through this, the reset is actually

m_deactivationTimer = 0;

since it's counting upwards instead of downwards. Maybe ResetTimer makes more sense. 

You can probably refactor it by placing ResetTimer in OnEnable() ,

Assets/_PartyParkade/Script/PowerUp/TextAndDrive.cs
```diff
@@ -0,0 +1,23 @@
+﻿using System.Collections;
```
m4rtint, line 1: Header

Assets/_PartyParkade/Script/PowerUp/BasePowerUp/Deployable.cs
m4rtint: Ohhhhh I see, so you're forcing the time to go over. 
Okay then maybe CancelTimer makes more sense.

wbeltran777: Setting m_deactivationTimer to 0 would skip over disabling the powerup though, which is what we want. 
Maybe calling the method EndTimer() is more accurate, since we have a deployable who's lifespan is cut short when someone grabs it. 
Not entirely sure what you mean by adding ResetTimer in OnEnable

m4rtint: I misunderstood the first message. 
EndTimer is good.

Assets/_PartyParkade/Script/PowerUp/TextAndDrive.cs
```diff
@@ -0,0 +1,29 @@
+﻿//-----------------------------------------------------------------------
+// <copyright file="PowerUpsObjectPooler.cs" company="PERIGON GAMES">
```
m4rtint, line 2: PowerUpsObjectPooler.cs -> TextAndDrive , if u want to change the small details

wbeltran777: Lol oops

mikeslgoh:
[(2) Create prefab for powerup](https://trello.com/c/cJmbDu94/614-2-create-prefab-for-powerup)

mikeslgoh:
[(2) Create deployable item](https://trello.com/c/XiSDolyT/615-2-create-deployable-item)

PR #243, opened 2020-06-15
Update Winterland.unity

m4rtint:
- Set the Camera Cull Mask Field to only be the the main objects and not everything. 
- Tag the Camera as MainCamera


mikeslgoh:
[(0.5) Fourth screen on three-player game freezes in corner map](https://trello.com/c/hFhxR16S/626-05-fourth-screen-on-three-player-game-freezes-in-corner-map)

PR #247, opened 2020-06-16
Removed BGM from the audio mixer

DJSteins:
Removed the BGM group from the audio mixer so that background music volume is solely controlled by the settings slider from the options panel.


Assets/_PartyParkade/Script/Data/SoundEffects.cs
```diff
@@ -16,7 +16,6 @@ public enum AUDIOGROUPS
         SFX,
         CarEngine,
         Boost,
-        BGM,
```
Gursaanj, line 19: This might be a small task for later, but we should rename this file, since its no longer just SoundEffects

m4rtint: But these are only Sound effects?

mikeslgoh:
[(1) Effects volume and master volume confusion](https://trello.com/c/WbeqVwOW/629-1-effects-volume-and-master-volume-confusion)

Assets/_PartyParkade/Script/Data/SoundEffects.cs
DJSteins: without bgm it is just sound effects

PR #248, opened 2020-06-16
Feature/adjusting points

DJSteins:
Adds a basic class for saving and loading game scores from player prefs and interfaces it with the ScoreManager.


Assets/_PartyParkade/Script/UI/ScoreSelectionBehaviour.cs
```diff
@@ -0,0 +1,35 @@
...
+
+    /// <summary>
+    /// Saves the given game score to PlayerPrefs and updates the ScoreManager with the new score.
+    /// This should be called when the player updates the score for the game.
+    /// </summary>
+    /// <param name="gameScore"></param>
+    public static void SaveGameScore(int gameScore)
+    {
+        PlayerPrefs.SetInt(PlayerPrefKeys.WinningScore, gameScore);
+        ScoreManager.WinningScore = gameScore;
```
m4rtint, line 33: Save it in GameData

And then on the WorldManager set up the ScoreManager with the GameData

DJSteins: Have a look now. Is that what you meant?

Assets/_PartyParkade/Script/Managers/WorldManagers/BaseWorldManager.cs
```diff
@@ -55,6 +55,7 @@ private void Start()
         m_playerManager.Initialize();
         m_parkingManager.Initialize();
         ScoreManager.ScoreSetup(PlayerManager.Players);
+        ScoreManager.WinningScore = GameData.WinningScore;
```
m4rtint, line 58: Yeah this is what i meant. 

Is it possible to pass in the winningscore through the "ScoreSetup"

DJSteins: done

Assets/_PartyParkade/Script/UI/ScoreSelectionBehaviour.cs
```diff
@@ -0,0 +1,35 @@
+﻿//-----------------------------------------------------------------------
+// <copyright file="ScoreSelectionBehaviour.cs" company="PERIGON GAMES">
+//     Copyright (c) PERIGON GAMES. 2019 All Rights Reserved
+// </copyright>
+//-----------------------------------------------------------------------
+using _PartyParkade.Script.Data;
+using UnityEngine;
+
+public class ScoreSelectionBehaviour : MonoBehaviour
```
m4rtint, line 9: We won't need the monobehaviour. 

Also remove the static from the methods.
Better habit not to use static too much for testing purpose.

DJSteins: Don't we want monobehaviour so we can use serializefield for UI objects when those get implemented? And I made the methods static in case the update behaviour is called somewhere else they can call the methods without needing to create the object

mikeslgoh:
[(1) Create Adjusting Points Implementation](https://trello.com/c/Hbfk7Zx7/642-1-create-adjusting-points-implementation)

Assets/_PartyParkade/Script/UI/ScoreSelectionBehaviour.cs
m4rtint: Oh i thought this would be a pure logic class, then sure, can stick with monobehaviour.

And so far, we don't need to access the methods from else where yet, so better to not use static for now since static makes it so that other classes would be tightly coupled to this class, and will crash if ScoreSelectionBehaviour does not exist in those places. 

The way i did it with the options panel is to create a new optionPanel (non mono) object to access the PlayerPrefs when needed. 
And then the optionPanelBehiovur just creates a new optionsPanel Object to access the playerprefs.

DJSteins: The reason I think it's okay for them to be static is that the methods don't modify the class itself. In a sense, if C# did not require classes, then these static methods would actually just be written as functions in a classless file. So just these static methods that don't affect the class itself are really not coupled to the class at all - they're only coupled to PlayerPrefs, PlayerPrefKeys, GameData, and ScoreManager, which are also all static.

As for separating logic for monobehaviour, I will put the logic in a separate file and then leave the monobehaviour to be filled in with UI stuff when we get there.

Assets/_PartyParkade/Script/UI/ScoreSelection.cs
```diff
@@ -0,0 +1,31 @@
+﻿//-----------------------------------------------------------------------
+// <copyright file="ScoreSelection.cs" company="PERIGON GAMES">
+//     Copyright (c) PERIGON GAMES. 2019 All Rights Reserved
+// </copyright>
+//-----------------------------------------------------------------------
+using _PartyParkade.Script.Data;
+using UnityEngine;
+
+public class ScoreSelection : MonoBehaviour
```
m4rtint, line 9: You can remove the MonoBehaviour here.

Assets/_PartyParkade/Script/UI/ScoreSelectionBehaviour.cs
m4rtint: That's fair.

Normally I was referring to tight coupling for the classes that calls the static method.  Like if MapSelection refers to calling the static method, it wouldn't be able to mock the ScoreSelection in test.

DJSteins: Oh okay I see what you mean now. If it isn't static then we can make classes that use ScoreSelection depend on it and, from a testing perspective, someone could make a subclass of ScoreSelection, or even a higher level interface, and implement methods that don't save to PlayerPrefs - similar to that example you sent in slack where we create a new database that doesn't actually charge the credit card.

m4rtint: yeah :).  
The annoying/hard part in these cases is because this is Unity and it's not entirely OOP, the general guidelines and rules can get muddy, but we try to work around it and do the best we can.

DJSteins: Updated to be not static now

PR #250, opened 2020-06-17
Feature/traffic light count down

m4rtint:
I didn't break down the method because I'm thinking it might be easier readability.


Assets/_PartyParkade/Script/UI/GameStartCountDown.cs
```diff
@@ -5,81 +5,86 @@
...
-        {
-            m_currentImage.sprite = m_greenTrafficLightSprite;
-            m_greenLightOn = true;
-        }
-        if (m_timer > 4.0 )
+        trafficSequence = DOTween.Sequence();
+        trafficSequence.PrependInterval(BeginningDelayTime);
+
+        trafficSequence.Append(m_redTrafficLight.DOFade(1, GlowAnimationTime));
+        trafficSequence.InsertCallback(RedLightSFXDelayTiming(), TrafficLightBeepSFX);
```
m4rtint, line 38: The callback insert a method to call at a specific time  - Thus the RedLightSFXDelayTiming

So if RedLightSFXDelayTiming = 1, then TrafficLightBeepSFX is called at 1second into the Sequence of animation

Assets/_PartyParkade/Script/UI/GameStartCountDown.cs
```diff
@@ -5,81 +5,86 @@
...
+    }
+
+    private float RedLightSFXDelayTiming()
+    {
+        return BeginningDelayTime + (GlowAnimationTime * 0.25f);
+    }
+
+    private float YellowLightSFXDelayTiming()
+    {
+        return BeginningDelayTime + GlowAnimationTime * 2 + (GlowAnimationTime * 0.25f);
```
m4rtint, line 74: The 0.25f is because I'm thinking the sound effect would play not at the start of the light, but in between light alpha = 0 and light alpha = 1, at 0.25% of when tween between 0 -> 1 alpha

mikeslgoh:
[End Game Message transition (1.0)](https://trello.com/c/60FdHJbg/653-end-game-message-transition-10)

mikeslgoh:
[(0.5) Update countdown timer based on new design](https://trello.com/c/jbyPNm4o/660-05-update-countdown-timer-based-on-new-design)

Assets/_PartyParkade/Script/UI/GameStartCountDown.cs
```diff
@@ -4,82 +4,84 @@
...
     {
-        Reset();
+        //TODO :
+        Debug.Log("Yellow / RED");
+        AudioManager.PlaySFX(AudioManager.PowerUpsSFX.CollectPowerup);
+    }
+
+    private void StartGameBeepSFX()
+    {
+        //TODO :
```
mikeslgoh, line 59: any reason why there are still TODO comments?

Assets/_PartyParkade/Script/UI/GameStartCountDown.cs
```diff
@@ -4,82 +4,84 @@
...
+        });
     }
 
-    private void OnEnable()
+    private void TrafficLightBeepSFX()
     {
-        Reset();
+        //TODO :
+        Debug.Log("Yellow / RED");
+        AudioManager.PlaySFX(AudioManager.PowerUpsSFX.CollectPowerup);
```
mikeslgoh, line 54: Should we make a task to find audio for this too, rather than using the powerup or parked sound effects?

Assets/_PartyParkade/Script/UI/GameStartCountDown.cs
m4rtint: It's for finding audio for this

Assets/_PartyParkade/Script/UI/GameStartCountDown.cs
m4rtint: yeah

Assets/_PartyParkade/Script/UI/GameStartCountDown.cs
```diff
@@ -4,82 +4,84 @@
...
-        }
+        });
     }
 
-    private void OnEnable()
+    private void TrafficLightBeepSFX()
     {
-        Reset();
+        //TODO :
+        Debug.Log("Yellow / RED");
```
wbeltran777, line 53: Probably wanna remove the debug logs

Assets/_PartyParkade/Script/UI/GameStartCountDown.cs
wbeltran777: Might be worth saving it as a const "m_audioOffset" or something like that, just in case we want to tweak it when we add the sound

m4rtint: Okay i added that constant

Assets/_PartyParkade/Script/UI/GameStartCountDown.cs
m4rtint: removed the debug logs

Added a bunch of comments, hopefully they make sense

Assets/_PartyParkade/Script/UI/GameStartCountDown.cs
```diff
@@ -4,82 +4,89 @@
...
-    private bool m_redLightOn = false;
-    private bool m_yellowLightOn = false;
-    private bool m_greenLightOn = false;
-    private bool m_isCompleted = false;
-    private Image m_currentImage = null;
-
-    private void Awake()
+    private const float GlowAnimationTime = 0.75f;
+    private const float BeginningDelayTime = 1.5f;
+    private const float AudioSFXDelay = 0.25f;
```
Gursaanj, line 18: should be using m_

m4rtint: Okay I'm confused. I thought const is no m_
 But public variable is m_

m4rtint: Or is it vice versa

Public static is no m_
Const is m_

Gursaanj: I thought anything that is private is m_ regardless of static or no static. and anything public is no m_ regardless static or no static?

m4rtint: https://stackoverflow.com/questions/242534/c-sharp-naming-convention-for-constants

Pascal for constants

PR #251, opened 2020-06-18
Bugfix/switch players

DJSteins:
- Changes the logic and model for switch players to try and prevent cars from going through the ground or the roofs of buildings, while still having the particle effects play at the correct time.
- Made it so that if all players are currently being tossed through the air, then someone else cannot activate their switch players powerup. There must be at least one other player who is not currently being switched for a player to use this powerup.
- Removed rotation and velocity reset from SetCarPosition() because SetCarPosition() is intended only for setting position. Made other functions for resetting rotation and velocity.
- Added some extra gravity acceleration logic so that cars feel heavier.


Assets/_PartyParkade/Script/PowerUp/SwitchAllPlayers.cs
```diff
@@ -221,16 +234,37 @@ private void FisherYatesShuffle(int n, List<PlayerController> controllers)
...
-            return;
+            return false;
         }
-        List<PlayerController> controllers = PlayerManager.Players.Where(x => x.CurrentPlayerNumber != playerController.CurrentPlayerNumber).ToList();
+        List<PlayerController> controllers = PlayerManager.Players.Where(x => IsValidPlayerForSwitch(playerController, x)).ToList();
         int n = controllers.Count;
+
+        // no available players to call switch on, so we return false to let the Activate method know that it
+        // should not activate this powerup. Prevents player from using the powerup when it would affect no one.
+        if (n == 0)
```
m4rtint, line 264: can you put this case into the switch?

Assets/_PartyParkade/Script/PowerUp/BasePowerUp/AbstractSwitchPlayers.cs
```diff
@@ -1,16 +1,22 @@
-
-
-
+//-----------------------------------------------------------------------
+// <copyright file="AbstractSwitchPlayers.cs" company="PERIGON GAMES">
+//     Copyright (c) PERIGON GAMES. 2019 All Rights Reserved
+// </copyright>
+//-----------------------------------------------------------------------
```
m4rtint, line 5: Do we still need an abstract switch player class? or just make one concrete?

DJSteins: Probably don't need it

Assets/_PartyParkade/Script/PowerUp/SwitchAllPlayers.cs
DJSteins: ye

DJSteins:
I added one more thing for some downwards acceleration to be added to the CarController so it feels like it has some more weight to it.

mikeslgoh:
[One car goes through the ground when all three players activate switch player powerup](https://trello.com/c/rCkUNTJD/631-one-car-goes-through-the-ground-when-all-three-players-activate-switch-player-powerup)

PR #253, opened 2020-06-19
Feature/text and drive conversation creator

Gursaanj:
Created A Converstation Creator Tool for the Text and Drive feature.

This tool lets you make and save a Conversation (Scriptable Object) into the project that can be read to get the appropriate messages.

Each message contains  A piece of text as well as a sprite (for allowing images within the text). Each text can hold Emojis as well (as seen in the gif below). where the Emojis will be displayed as such to the user but will in be rich text form (the form needed by textmeshpro to be displayed)  when saving it in unity.

ConversationCreator


m4rtint:
I don't know how necassry this is, but as a backlog, u can add a "choose your directive" to save in sort of thing.

Assets/_PartyParkade/Script/PowerUp/DriveText/Conversation.cs
```diff
@@ -0,0 +1,55 @@
...
+    /// <summary>
+    /// the list of messages associated to conversation
+    /// </summary>
+    [SerializeField]
+    public List<TextAndDriveMessages> Messages;
+}
+
+
+[System.Serializable]
+public struct TextAndDriveMessages
```
m4rtint, line 26: For naming conventions, because this is ONE message, then it should be just TextAndDriveMessage(without the s in the end)

Assets/_PartyParkade/Script/PowerUp/DriveText/ConversationCreatorLogic.cs
```diff
@@ -0,0 +1,180 @@
...
+    /// Static method to get the list of sprites (emojis) that can be used in a piece of text
+    /// set up to initialize if null;
+    /// </summary>
+    public static string[] InlineSpriteNames
+    {
+        get
+        {
+            if (m_inlineSpriteNames == null)
+            {
+                Sprite[] sprites = Resources.LoadAll<Sprite>("EmojiSprites");
```
m4rtint, line 60: So this is assuming that the emojis we get is a spritesheet that we cut up?

Gursaanj:
[(4) Scriptable object for messages](https://trello.com/c/GbweqAAj/616-4-scriptable-object-for-messages)

Gursaanj:
[(4) Scriptable object for messages](https://trello.com/c/GbweqAAj/616-4-scriptable-object-for-messages)

Gursaanj:
[(4) Scriptable object for messages](https://trello.com/c/GbweqAAj/616-4-scriptable-object-for-messages)

Assets/_PartyParkade/Script/PowerUp/DriveText/ConversationCreatorLogic.cs
```diff
@@ -0,0 +1,180 @@
...
+    }
+
+    /// <summary>
+    /// Decides if the current message can be switched with the message above it
+    /// </summary>
+    /// <param name="indexOfMessage"></param>
+    /// <returns></returns>
+    public bool CanMessageBeIncremented(int indexOfMessage)
+    {
+        return m_listOfMessageLogic != null && m_listOfMessageLogic.Count > 1 && indexOfMessage != 0;
```
m4rtint, line 100: So I realized something recently
Because this list is public -> you set up a property

So because you have a property setup, you can make the getter create a new List<> if it's null

So within this class, you don't need to null check and call that properties getter, so it'll never be null.

Assets/_PartyParkade/Script/PowerUp/DriveText/ConversationCreatorLogic.cs
```diff
@@ -0,0 +1,180 @@
...
+
+    /// <summary>
+    /// Finds out if conditions are met to allow users to save conversation
+    /// </summary>
+    /// <returns></returns>
+    public bool CanConversationBeSaved()
+    {
+        if (string.IsNullOrWhiteSpace(Title))
+        {
+            m_cantSaveErrorMessage = "This conversation needs a title";
```
m4rtint, line 135: You can store them as const if u want

Assets/_PartyParkade/Script/PowerUp/DriveText/ConversationCreatorLogic.cs
```diff
@@ -0,0 +1,180 @@
...
+
+        return true;
+    }
+
+    /// <summary>
+    /// Creates Conversation and saves it within the described path
+    /// </summary>
+    public void SaveConversation()
+    {
+        Conversation conversation = (Conversation)ScriptableObject.CreateInstance("Conversation");
```
m4rtint, line 162: Can you use the Type Conversation instead of the String?

Assets/_PartyParkade/Script/PowerUp/DriveText/ConversationCreatorLogic.cs
Gursaanj: Yeah, for the emojis to be called properly by Textmeshpro the easiest way to do that would be to use a spritesheet anyway

Gursaanj:
Allowed for people to see the inline sprites (emojis) when they select them from the dropdown

image

Assets/_PartyParkade/Script/PowerUp/DriveText/ConversationCreatorLogic.cs
```diff
@@ -60,21 +60,27 @@ public List<MessageLogic> ListOfMessageLogic
...
     /// </summary>
-    public static string[] InlineSpriteNames
+    public static GUIContent[] InlineSpriteContents
     {
         get
         {
-            if (m_inlineSpriteNames == null)
+            if (m_inlineSpriteContents == null)
             {
                 Sprite[] sprites = Resources.LoadAll<Sprite>("EmojiSprites");
```
m4rtint, line 69: Because you're finding a file with a specific string name, might actually be better to try catch this, or some sort of log that tells you 

YO, I'M LOOKING FOR A FILE NAME CALLED EMOJI SPRITES, some sort of check that prints out if it can find this specific file with this specific name.

Gursaanj: Good point, planning on changing this sooner or later so people can decided on which sprite should be used for the conversation creator tool (defaulting to "EmojiSprites'). But for now, a log will do it

m4rtint: yeah so long term it'll be a place where you pick which file you want to import.

PR #255, opened 2020-06-22
BugFix/Master Slider and BGM Slider

DJSteins:
Needed to change the values being passed into the update audio functions so that bgm doesn't override master volume


DJSteins:
[(1) Effects volume and master volume confusion](https://trello.com/c/WbeqVwOW/629-1-effects-volume-and-master-volume-confusion)

PR #256, opened 2020-06-22
Feature/text and drive header

wbeltran777:
- Created Text and Drive prefab. Can be activated by grabbing deployable. Will spawn the text and drive canvas. Nothing destroys it at the moment so it'll just stay on when grabbed. 
- Added Get Camera method to PlayerController. Makes it easy to assign camera when collecting powerup.


m4rtint:
macbook11
macbook13
144p
1080p


The Powerup should be under the header. 
Just make sure the anchors you place are the correct percentage. 
e.g. - header: 0 -> 0.1
 - power up  0.12 -> 0.20

Something like that. 

Also the header should be over the border.

m4rtint:
The Power Up Object Pooler Prefab does not have text and drive pushed
So it'll cause problems in the powerup debugger.

Assets/_PartyParkade/Script/PowerUp/TextAndDrive.cs
```diff
@@ -12,18 +12,19 @@ namespace _PartyParkade.Script.Game
...
+            // With order being 1, the text and drive UI will lie on top of everything else in view
+            m_textAndDriveHeader.sortingOrder = 1;
 
-        /// <summary>
-        /// Handles filling the player's screen with text messages
-        /// </summary>
-        private void TextAndDriveTriggered()
-        {
-            // TODO
+            Canvas activeHeader = Instantiate(m_textAndDriveHeader);
```
m4rtint, line 26: This shouldn't be instantiating the TextAndDrive. 

The TextAndDrive GameObject should be within the player's object. 

The Player UserInterfaceView or another script should turn this On/Off.

Assets/_PartyParkade/Script/PowerUp/TextAndDrive.cs
```diff
@@ -12,18 +12,19 @@ namespace _PartyParkade.Script.Game
...
+
         protected override void HandleTrigger(Collider other)
         {
-            TextAndDriveTriggered();
-            EndTimer();
-        }
+            m_textAndDriveHeader.worldCamera = other.GetComponentInParent<PlayerController>().GetCamera();
+            m_textAndDriveHeader.planeDistance = m_planeDistance;
+            // With order being 1, the text and drive UI will lie on top of everything else in view
+            m_textAndDriveHeader.sortingOrder = 1;
```
m4rtint, line 24: If you put the gameobject within the PlayerUserInterface Prefab, I don't think you'll need to do all this.

Assets/_PartyParkade/Script/PowerUp/TextAndDrive.cs
wbeltran777: I'll get on this fix , should fix the power up box hiding behind the border issue

wbeltran777: Would it be reasonable to have the Player Controller turn on the text and drive component? 
It's easily accessible when a car collides with a powerup.

m4rtint: yup that's best

mikeslgoh:
[Create UI header (1.0)](https://trello.com/c/GlKgrIqd/651-create-ui-header-10)

PR #257, opened 2020-06-23
Character Confirm Button Refactor

m4rtint:
- Added new method to handle different input for different states


m4rtint:
[Player selection navigation for colors/hats (0.5)](https://trello.com/c/BLZDFe1L/666-player-selection-navigation-for-colors-hats-05)

PR #258, opened 2020-06-23
Ensure that the TextAndDriveConversationCreator will not break build

Gursaanj:


Gursaanj:
> If build fails, I would recommend only settings the if statement around the AssetDatabase stuff

Normally I would agree but in this case, since all the logic for both these classes will never and can never be used in the build, it would just save memory/space if we just negated the entire script

PR #259, opened 2020-06-24
Feature/adjust points ui

DJSteins:
UI now includes implementation for adjusting game score


Assets/_PartyParkade/Script/UI/ScoreSelectionBehaviour.cs
```diff
@@ -3,20 +3,201 @@
...
+            }
+        }
+    }
+
+    /// <summary>
+    /// Visualize the text for the game score
+    /// </summary>
+    private void SetText()
+    {
+        string baseText = m_score == 1 ? "Point" : "Points";
```
m4rtint, line 196: You can use MainText.PointSingular/PointPlural, it's already been made instead of the strings here

Assets/_PartyParkade/Script/UI/ScoreSelectionBehaviour.cs
```diff
@@ -3,20 +3,201 @@
...
+        {
+            m_rightButtonSprite = m_rightButton.image.sprite;
+        }
+        if (m_leftButton)
+        {
+            m_leftButtonSprite = m_leftButton.image.sprite;
+        }
+
         m_scoreSelection = new ScoreSelection();
+        if (m_scoreSelection != null)
```
m4rtint, line 159: so for the null check, it's up to you but what do you think about putting scoreSelection in a property so you don't have to null check everywhere?

mikeslgoh:
[Add the UI for the adjusting points panel on Map Selection Screen (2.0)](https://trello.com/c/SQdvvrDD/663-add-the-ui-for-the-adjusting-points-panel-on-map-selection-screen-20)

Assets/_PartyParkade/Script/UI/ScoreSelectionBehaviour.cs
```diff
@@ -3,20 +3,201 @@
...
+
+    /// <summary>
+    /// Visualize the text for the game score
+    /// </summary>
+    private void SetText()
+    {
+        string baseText = m_score == 1 ? "Point" : "Points";
+        if (m_pointsText)
+        {
+            m_pointsText.SetText(m_score + " " + baseText);
```
Gursaanj, line 199: You can use string.Format here

Assets/_PartyParkade/Script/UI/ScoreSelectionBehaviour.cs
DJSteins: Isn't it already a property?

m4rtint: so in terms of C# terminology

Field: int something = 1

Property: int Something { get; set}

PR #263, opened 2020-06-25
BugFix Activate Input

m4rtint:
Sometimes when you press activate, it doesn't register. Swapped to OnUpdate.


m4rtint:
[Bug: Activate sometimes doesn't register (0.5)](https://trello.com/c/wpHjWauF/677-bug-activate-sometimes-doesnt-register-05)

PR #265, opened 2020-06-30
Feature/residential redesign

wbeltran777:
- Residential work in progress will be adding more stuff in future PRs but here's a good spot to update it. 
- Street parking prefab was modified because it didn't stash the particle system in its child
- Added trees to project, future PR will be adding them onto the map itself
- Added all the new roads
- Future PR will add more of the new buildings and add them to Residential if relevant
Residential


mikeslgoh:
[(3) Re-design the Residential map using Magan 3D models](https://trello.com/c/Ceykb29Z/408-3-re-design-the-residential-map-using-magan-3d-models)

wbeltran777:
@m4rtint  Is that all the trees files from the patreon you want to delete, or all the trees we had from the original residential asset?

m4rtint:
> @m4rtint  Is that all the trees files from the patreon you want to delete, or all the trees we had from the original residential asset?

From original

PR #266, opened 2020-07-01
Feature/text and drive conversation preview

Gursaanj:
Created a Preview tool for the conversation creator. So people can see their conversations in as they would appear (sort of) in the game. To do this I created a specific scene, where if the conversation creator is used will allow the preview to take place. 

Since the conversation Behavior will only in Play mode, I created the rule that the preview capability will only work in Play mode. 

A difficult situation came up in which if you were to begin a conversation in Edit mode and wanted to preview it, and then go to PlayMode to do so, all your work for the conversation would be overridden and go back to null (something to do with non serialized data unable to be serialized for non Monobehaviour scripts). 

The solution here (as suggested in Unity Forums) was to save a temporary Conversation scriptableObject right before you enter PlayMode and then load it back in a few frames after entering PlayMode, then unload the asset alltogether. 

Not the cleanest solution, but it works and will allow us to create conversations more easily. Might Update solution in a later date if time permits (most likely using this interface [ISerializationCallbackReceiver
](https://docs.unity3d.com/ScriptReference/ISerializationCallbackReceiver.html) )

ConversationCreator Preview


Assets/Editor/ConversationCreator.cs
```diff
@@ -318,4 +329,63 @@ private MessageLogic DrawMessageBox(MessageLogic messageLogic, int indexOfMessag
 
         return messageLogic;
     }
+
+    /// <summary>
+    /// Checks if there is any Temporary Conversation (made between PlayStateModes) and sets
+    /// the data appropriately
+    /// </summary>
+    private void CheckForTemporaryConversation()
```
Gursaanj, line 337: Unfortunately, had to keep the logic here, instead of ConversationCreatorLogic.cs. Cant say exactly why it wasnt working but when I Placed the CreateTemporaryConversation Method into the Logic script, the delay calls in this method would never call

Gursaanj:
[Add the UI for the adjusting points panel on Map Selection Screen (2.0)](https://trello.com/c/SQdvvrDD/663-add-the-ui-for-the-adjusting-points-panel-on-map-selection-screen-20)

Gursaanj:
[Preview Tool (3.0)](https://trello.com/c/OszZ9pqn/665-preview-tool-30)

Assets/_PartyParkade/Script/PowerUp/DriveText/ConversationCreatorLogic.cs
```diff
@@ -7,17 +7,23 @@
 using System.Collections.Generic;
 using UnityEditor;
 using UnityEngine;
+using UnityEngine.SceneManagement;
 
 public class ConversationCreatorLogic
```
m4rtint, line 12: Shouldnt' this be placed in the Tools folder?

Assets/_PartyParkade/Script/PowerUp/DriveText/ConversationCreatorLogic.cs
```diff
@@ -147,6 +154,27 @@ public MessageLogic SwitchMessages(MessageLogic currentMessageLogic, int targetM
         return currentMessageLogic;
     }
 
+    /// <summary>
+    /// Finds out if conditions are met to allow users to preview conversations
+    /// </summary>
+    /// <returns></returns>
+    public bool CanConversationBePreviewed()
+    {
+        if (!SceneManager.GetActiveScene().name.Equals("TextAndDrivePreview"))
```
wbeltran777, line 163: Should we add this text to the string public class?

Gursaanj: Good point

m4rtint:
Conflicts - the only thing I added for the Conversation.prefab is a canvasgroup i think

Assets/_PartyParkade/Script/PowerUp/DriveText/ConversationCreatorLogic.cs
m4rtint: I think you can just put this on the top of this script since this is only a editor tool

PR #268, opened 2020-07-02
Removed old prefabs for the new buildings with texture maps

mikeslgoh:
Also added in the new toolbox, update the size of the powerups on each map

NOTE: Residential will have missing buildings that I cannot edit since it's part of the prefab


mikeslgoh:
[Import Magan's texture maps into the project (0.5)](https://trello.com/c/cgiqfzNR/686-import-magans-texture-maps-into-the-project-05)

PR #276, opened 2020-07-06
Feature/colliders in winter land

wbeltran777:
- Added colliders to winterland so now collisions don't depend on the terrain
- Saved winterland geometry as prefab
- Modified current road prefabs to disable their colliders


mikeslgoh:
[Update colliders on roads for Winterland (1.0)](https://trello.com/c/9s0NJVoB/638-update-colliders-on-roads-for-winterland-10)

m4rtint:
Where does the conflict come from?

wbeltran777:
> Where does the conflict come from?

Modifying the terrain and adding the colliders

PR #277, opened 2020-07-06
Added CarSphere to the Car Layer

m4rtint:
- Oil Spill only collides with the carsphere layer
- Added carsphere layer to the car


mikeslgoh:
[Use particle effect for oil spill (2.0)](https://trello.com/c/ZHPExr9x/683-use-particle-effect-for-oil-spill-20)

PR #280, opened 2020-07-06
Cull Sphere bug fix

wbeltran777:
Fixed layer masking bug, occlusion should now shrink when car is not in front of an Obstacle


m4rtint:
I see you added 2 materials to the car, how does that work with the cull sphere?

mikeslgoh:
[Cull sphere fix](https://trello.com/c/0hXymdPg/694-cull-sphere-fix)

wbeltran777:
@m4rtint The CameraCulling script has a serialized layer mask value. Before it was ignoring everything, which caused the cullsphere to always be on. 
Right now it's set so that the raycast only looks at Obstacles or the PlayerModel. So if it collides with the player model, the cullsphere doesn't expand, and if it collides with an obstacle then it does expand.

PR #281, opened 2020-07-07
Enhancment/parking spot selection

DJSteins:
Implements the parking spot queue


mikeslgoh:
[Parking spot selection (1.0)](https://trello.com/c/8u2Wo0pd/624-parking-spot-selection-10)

Assets/_PartyParkade/Tests/ParkingSpotQueue_Tests.cs
```diff
@@ -0,0 +1,128 @@
...
+
+        [Test]
+        public void GenerateParkingSpotOrder_ZeroMinDistanceSelectAll()
+        {
+            ParkingSpotQueue queue = new ParkingSpotQueue(0);
+
+            queue.GenerateParkingSpotOrder(m_basicPositions, m_basicPositions.Count);
+
+            List<Vector3> queuedPositions = GetAllQueuedPositions(queue);
+            Assert.AreEqual(queuedPositions.Count, m_basicPositions.Count);
```
m4rtint, line 41: Can you seperate the Count as variables and name them
"ExpectedResult"
and
"Actual Results"

normally you hard code the expected result and make sure what you get is = the hard coded result

Assets/_PartyParkade/Tests/ParkingSpotQueue_Tests.cs
```diff
@@ -0,0 +1,128 @@
...
+        private List<Vector3> m_basicPositions = new List<Vector3>
+        {
+            new Vector3(0, 0, 0),
+            new Vector3(1, 0, 0),
+            new Vector3(1, 0, 1),
+            new Vector3(0, 0, 1)
+        };
+
+        [SetUp]
+        public void Setup()
```
m4rtint, line 24: You can remove these if u're not using these

Assets/_PartyParkade/Tests/ParkingSpotQueue_Tests.cs
```diff
@@ -0,0 +1,128 @@
...
+
+        [Test]
+        public void GenerateParkingSpotOrder_QueueSizeGreaterThanSpotCount()
+        {
+            ParkingSpotQueue queue = new ParkingSpotQueue(0);
+
+            queue.GenerateParkingSpotOrder(m_basicPositions, m_basicPositions.Count + 2);
+
+            List<Vector3> queuedPositions = GetAllQueuedPositions(queue);
+            Assert.AreEqual(queuedPositions.Count, m_basicPositions.Count + 2);
```
m4rtint, line 81: yeah I just realized, it's better to hard code the "expected value" cuz if someone else comes along and tries to fix it, they won't know what value they will be aiming for because of the two variables. 
Or at least it's easier to visualize if we ever need to fix this.

PR #293, opened 2020-07-13
Bugfix/final parking spot doesnt activate

mikeslgoh:
The fix is that if there are more than 1P, the queue size would be:

**1P with winning score + (other players * (winning score - 1))** and if it is only 1P then the queue size is the **winning score**

since this was what we had talked about in the dev meeting


mikeslgoh:
[Final parking spot doesn't activate](https://trello.com/c/ujlGysBa/717-final-parking-spot-doesnt-activate)

Assets/_PartyParkade/Script/Managers/ParkingManager.cs
```diff
@@ -85,7 +85,18 @@ public void ActivateParkingSpot(GameObject parkingSpot)
     private void GenerateParkingSpotOrder()
     {
         List<Vector3> parkingSpotPositions = m_targets.Select(x => x.transform.position).ToList();
-        ParkingSpotQueue.GenerateParkingSpotOrder(parkingSpotPositions, (GameData.WinningScore *  GameData.PlayerCount) - 1);
+        int queueSize = 1;
```
m4rtint, line 88: All this stuff isn't used? 

Logic should all be inside GenerateParkingSpotOrder, the tests should also be updated.

Assets/_PartyParkade/Script/Managers/ParkingManager.cs
```diff
@@ -85,7 +85,18 @@ public void ActivateParkingSpot(GameObject parkingSpot)
...
+        int winningScore = GameData.WinningScore;
+        int count = GameData.PlayerCount;
+
+        if (GameData.PlayerCount > 1)
+        {
+            queueSize = winningScore + ((count-1) * (winningScore-1));
+        } else
+        {
+            queueSize = winningScore;
```
DJSteins, line 97: If the `GameData.PlayerCount == 1`, then the `((count - 1) * (winningScore-1))` will be zero anyways, so you don't need two cases here, just the first case.

i.e. `queueSize = winningScore + ((count-1) * (winningScore-1));` covers all possibilities

Assets/_PartyParkade/Script/Managers/ParkingManager.cs
```diff
@@ -85,7 +85,10 @@ public void ActivateParkingSpot(GameObject parkingSpot)
     private void GenerateParkingSpotOrder()
     {
         List<Vector3> parkingSpotPositions = m_targets.Select(x => x.transform.position).ToList();
-        ParkingSpotQueue.GenerateParkingSpotOrder(parkingSpotPositions, (GameData.WinningScore *  GameData.PlayerCount) - 1);
+        int winningScore = GameData.WinningScore;
+        int count = GameData.PlayerCount;
+        int queueSize = winningScore + ((count-1) * (winningScore-1));
+        ParkingSpotQueue.GenerateParkingSpotOrder(parkingSpotPositions, (GameData.WinningScore *  GameData.PlayerCount));
```
m4rtint, line 91: I think u forgot to replace this with queueSize.

Assets/_PartyParkade/Script/Managers/ParkingManager.cs
mikeslgoh: I don't know if we wrote tests for this @DJSteins and if want to move this logic into GenerateParkingSpotOrder, then we would also need to pass it GameData since it doesnt currently reference it (I just worked with what was already there)

m4rtint: Well since that logic has basically been removed, it's fine. 

But still need to update the tests to have 1 player and 1 point

mikeslgoh: We don't have tests for parking manager and in ParkingSpotQueue_Tests, it does not include the part I added so unless you want me to still right a test to check that a single parking spot is returned

m4rtint: yeah write the test for the single parking spot is returned inside ParkingSpotQueue_Tests

PR #294, opened 2020-07-13
Sfx/traffic light

DJSteins:
Add sfx for the traffic light countdown and move BGM starting to after the countdown


Assets/_PartyParkade/Script/Managers/AudioManager.cs
```diff
@@ -228,17 +231,17 @@ private void StartAmbientNoises(string sceneName)
             }
         }
 
-        private void PlaySceneSoundConnection(Scene scene, LoadSceneMode mode)
+        private void PlaySceneSoundConnection()
         {
             StopAmbientNoises();
             SetUpBackgroundMusicConnections();
-            if (scene.buildIndex != GENERIC_UI_BUILD_INDEX)
+            if (m_activeScene.buildIndex != GENERIC_UI_BUILD_INDEX && m_activeScene.buildIndex != -1)
```
m4rtint, line 238: when would it be -1?

DJSteins: I don't know the exact reason why it was happening, but for some reason `m_activeScene` was getting set to a scene with a `buildIndex` of `-1`, so I added that in for safe measure

DJSteins: But I guess now that I have added the extra check in the other setup function, I probably don't need to check for `-1` here

DJSteins: Actually for some reason this method gets called before the setup sometimes. I don't know why, but that's how we end up with `-1`

mikeslgoh:
[Add SFX for traffic light intro (0.5)](https://trello.com/c/AOzkAnTE/661-add-sfx-for-traffic-light-intro-05)

Assets/_PartyParkade/Script/Managers/AudioManager.cs
m4rtint: This is called on sceneload, so it's probably before most things are called. okay this is fine

PR #295, opened 2020-07-14
Feature/text and drive linking prefab

wbeltran777:
- Added text and drive model to the prefab. Rotates when it's active 
- Added two test conversations
- List in ConversationBehaviour can be populated with scriptable objects, a function will pick one at random when conversation starts. 
- Added StartConversation function for triggering conversation and reseting the state of conversation behaviour before displaying new messages


mikeslgoh:
[Link the prefab into the game (2.0)](https://trello.com/c/2zyXf86C/613-link-the-prefab-into-the-game-20)

Assets/_PartyParkade/Script/PowerUp/TextAndDrivePickUpBehaviour.cs
```diff
@@ -4,12 +4,22 @@
...
+using DG.Tweening;
 using UnityEngine;
 
 namespace _PartyParkade.Script.Game
 {
     public class TextAndDrivePickUpBehaviour : Deployable
     {
+        private void Awake()
+        {
+            float randomRange = Random.Range(90, 180);
```
m4rtint, line 16: Put this onEnable. 

And then Kill the Tween on Disable

Assets/_PartyParkade/Script/PowerUp/DriveText/ConversationBehaviour.cs
```diff
@@ -89,6 +92,20 @@ public void SetMessage()
                 AnimateConversation(0);
         }
 
+        /// <summary>
+        /// Pick a random Conversation and start displaying it
+        /// </summary>
+        public void StartConversation()
```
m4rtint, line 98: For sake of organization, i think it's best if you put the 
[SerializeField] m_listOfConversation inside the

TextAndDriveBehaviour.cs so it's more organized in a way that this Behaviour is stateless, and TextAndDriveBehaviour would have all the modules

So have the listOfConversation inside TextAndDriveBehaviour and randomize it there, and then pass it into here. 

and then to call it, you can use @Gursaanj method "SetAndPlayPreviewConversation()" and get rid of the #if UNITY_EDITOR stuff.

Assets/_PartyParkade/Script/PowerUp/DriveText/ConversationBehaviour.cs
```diff
@@ -128,6 +145,11 @@ public void SetAndPlayPreviewConversation(Conversation conversation)
         }
 #endif
 
+        private Conversation GetRandomConversation()
+        {
+                return m_listOfConversation[(int) (UnityEngine.Random.Range(0,m_listOfConversation.Count - 0.0001f))];
```
m4rtint, line 150: I'm pretty sure the Unity Random Range excludes the last value. No need the - 0.0001f, which normally is -1, not sure why u have a float there.

Assets/_PartyParkade/Script/PowerUp/DriveText/ConversationBehaviour.cs
m4rtint: Oh and rename SetAndPlayPreviewConversation to something more suiting.

Assets/_PartyParkade/Script/PowerUp/DriveText/ConversationBehaviour.cs
wbeltran777: Random range is inclusive though. So it can return m_listOfConversation.Count which would cause indexOutOfRange. 
I subtracted 0.001f so our max is val from the random truncates to (m_listOfConversation - 1)

m4rtint: ah okay it is inclusive, 

But that's kinda strange how it's 0.0001, why not 0.1, why not 0.001?

m4rtint: https://docs.unity3d.com/ScriptReference/Random.Range.html

Okay so Return a random integer number between min [inclusive] and max [exclusive] (Read Only).

as long as it's an int, which is should be, it'll be exclusive. When it's a float, it's inclusive.

wbeltran777: Well if it was (int) Random.Range(0,1.9), you have a higher chance of landing on a 0 than on a 1. 
Random.Range(0,1.9999) makes the odds a bit better. 
I figure four decimal places is enough of a magnitude that the bias towards all numbers that arent m_listOfConversation.Count - 1 is pretty minimal.

wbeltran777: Oh lol didn't see the int documentation, will change it

m4rtint: But if you do 
Random.Range(0, 2), then you''ll have an equal change of 0 and 1 right? 
since it can only randomly choose either 0 or 1?

wbeltran777: No yeah I deleted my explanation, the int randomizer should fix all woes

wbeltran777:
@m4rtint Yep yep! Grabbed two bells and no issues there

Assets/_PartyParkade/Script/PowerUp/TextAndDrivePickUpBehaviour.cs
```diff
@@ -4,12 +4,35 @@
...
+            transform.DORotate(new Vector3(0, 360, 0), 3.0f)
+                    .SetLoops(-1, LoopType.Incremental)
+                    .SetEase(Ease.Linear)
+                    .SetId(m_currentID);
+        }
+
+        protected override void OnDisable()
+        {
+            base.OnDisable();
+            DOTween.Kill(targetOrId: m_rotationID);
```
m4rtint, line 32: Instead of all the Id stuff, you can just call
transform.DOKill();

PR #296, opened 2020-07-14
Power up now spins on center axis, and all power up placements have b…

wbeltran777:
…een updated in residential and winterland

As the title says. Added a 'pivot' component in the tool box prefab so when the model spins it actually spins on it's center axis.

Also deleted all the level meshes since we don't need those anymore.


Assets/_PartyParkade/Script/Game/ItemPickUp.cs
```diff
@@ -20,6 +20,8 @@ public class ItemPickUp : MonoBehaviour
         [SerializeField]
         private MeshRenderer m_pickupMesh = null;
         [SerializeField]
+        private Transform m_meshPivot = null;
```
mikeslgoh, line 23: why did you need to add another variable?

m4rtint, line 23: So a work around instead of updating the mesh? lol

mikeslgoh: Did it not spin correctly with using just the mesh renderer?

m4rtint: I'm guessing because of the mesh center point is off, we're making our own. Honestly thinking about, i'm half sure that you can probably get the "parent" which is probably the mesh pivot point?

wbeltran777: Oh shoot I didn't realize there was another tool box mesh to replace it with

mikeslgoh:
[Powerup placement (1.0)](https://trello.com/c/9BkG6QNn/710-powerup-placement-10)

PR #297, opened 2020-07-14
Test/extract materials from models

mikeslgoh:
Extracting materials from Magan's 3D assets as it seems to fix the issue of invisible buildings for the Github builds


mikeslgoh:
[(2) No Building is shown in Wonderland](https://trello.com/c/imzzPaBh/587-2-no-building-is-shown-in-wonderland)

PR #303, opened 2020-07-18
BugFix - Character Selection

m4rtint:
- Character selection bug fix - used controller dead zone. 
 - joystick press right 

- destructible OnDestroy unsubscribes instead of ondisable - ( strange behaviour)


m4rtint:
[Controller doesnt work properly on player selection](https://trello.com/c/4pZmTuqB/719-controller-doesnt-work-properly-on-player-selection)

Assets/_PartyParkade/Script/Game/Destructables/Destructible.cs
```diff
@@ -71,7 +71,7 @@ private void Start()
             StateManager.Instance.OnPreGame += CleanUp;
         }
 
-        private void OnDisable()
+        private void OnDestroy()
```
mikeslgoh, line 74: So this was the reason the weird behaviour @wbeltran777 reported was happening?

m4rtint: yuh

m4rtint: Maybe OnDisable isn't called on scene change, but OnDestroy is called

PR #304, opened 2020-07-18
Clean Up Maps - Residential / Snow

m4rtint:
- Removed Random Cylinders from the Game Objects
- Zeroed out GameObjects that carried a collection of GameObjects. E.g. ParkingsSpots (which contained all the parking spots)
- PowerUpObjectPooler (SHOULD NOT HAVE CHILDREN), Object Poolers should NOT have children. It GENERATES ITS OWN CHILDREN, which is POOL OF OBJECTS
- Renamed PowerUp as PickUpPowerUp to avoid confusion, created a parent that carries them.


m4rtint:
[Go through and check for any empty/missing prefabs and objects (3.0)](https://trello.com/c/gbZoaXZF/697-go-through-and-check-for-any-empty-missing-prefabs-and-objects-30)

PR #305, opened 2020-07-18
Remove cache instruction from builds

mikeslgoh:
The builds failed due to caching so I'm removing it - tested and the builds are fine


m4rtint:
Why is caching causing problems?

mikeslgoh:
> Why is caching causing problems?

For some reason the build failed previously plus for our builds, we don't seem to use caching anyway.
image

m4rtint:
> > Why is caching causing problems?
> 
> For some reason the build failed previously plus for our builds, we don't seem to use caching anyway.
> image

Were you able to find out why it might be timing out?

mikeslgoh:
> > > Why is caching causing problems?
> > 
> > 
> > For some reason the build failed previously plus for our builds, we don't seem to use caching anyway.
> > image
> 
> Were you able to find out why it might be timing out?

@m4rtint no ... there isn't anything I could find that would be the cause

mikeslgoh:
Closing this PR because it looks like I just needed to re-run the build job

PR #307, opened 2020-07-19
Creative/text and drive

wbeltran777:
- Added the conversation we made for the social
- Also added quick fix for item pick up null bug


mikeslgoh:
[Text and Drive Brainstorm Session](https://trello.com/c/EBEI6ywL/722-text-and-drive-brainstorm-session)

PR #309, opened 2020-07-20
Added single  player world manager and manager picker

wbeltran777:
- Added WorldManagerPicker for picking between SinglePlayerManager and MultiplayerManager (used to be called DefaultWorldManager)
- Deleted TimeManager prefab. Logic being that it's only used in Single player, and since it doesn't require serialized fields, it feels cleaner to just add it as a component instead of a prefab.


Assets/_PartyParkade/Script/Managers/TimeManager.cs
```diff
@@ -62,8 +63,8 @@ private void TimerUpdate()
         {
             TimerUpdateEvent();
         }
-
-        m_currentGameTime += Time.deltaTime * m_CurrentTimeMultiplier;
+        // TimeSpan class is useful for formatting the time as minutes:seconds:milliseconds
```
wbeltran777, line 66: Left this is as a reminder for later

Assets/_PartyParkade/Script/Managers/WorldManagers/WorldManagerPicker.cs
```diff
@@ -0,0 +1,19 @@
+﻿using System.Collections;
+using System.Collections.Generic;
+using UnityEngine;
+using _PartyParkade.Script.Data;
+
+public class WorldManagerPicker : MonoBehaviour
+{
+    private void OnEnable()
```
m4rtint, line 8: Okay this is a good start. 
However, it's not going to be for Testing, you won't be able to change GameData.PlayerCount easily for testing. 

So you'll need to 
1) remove the serializedField of m_numberOfPlayers inside the PlayerManager.cs file. 
2) Place
`private void SetupPlayerCount()
    {
        if (GameData.IsStartingFromMenu)
        {
            m_numberOfPlayers = GameData.PlayerCount;
        }
    }`

Inside this class, and serialize the m_numberOfPlayers inside this class. So from now, It's the WorldManagerPicker that chooses how many players. This function is also inside the PlayerManager class, you can remove it from there.

3) Inject/Pass in the m_numberOfPlayers into the WorldManager (You can use the abstract class). Which then injects that number into the PlayerManager (There should be Initialize() methods you can pass it through there.

This way you can change how many players are created and it should setup everything correctly.

wbeltran777: Sounds good! Will get right on that

Assets/_PartyParkade/Script/Managers/WorldManagers/WorldManagerPicker.cs
```diff
@@ -0,0 +1,19 @@
+﻿using System.Collections;
```
Gursaanj, line 1: Missing Header

Assets/_PartyParkade/Script/Managers/WorldManagers/BaseWorldManager.cs
```diff
@@ -19,6 +19,15 @@ public abstract class BaseWorldManager : MonoBehaviour
     private PlayerManager m_playerManager;
     private ParkingManager m_parkingManager;
 
+    private int m_numberOfPlayers;
+    public int PlayerNum
```
m4rtint, line 23: Name this the proper PlayerNumber

m4rtint:
image
I'm not sure if you're supposed to overwrite or revert ,but the Manager has some stuff here that's left specifically for this map?  I think it's better if you press the override button.

mikeslgoh:
[Create timemanager (3.0)](https://trello.com/c/Ah9QGwcj/728-create-timemanager-30)

PR #310, opened 2020-07-21
Bugfix - Main menu BGM does not play

mikeslgoh:
For some reason after updating the SetupSceneSoundConnection to match the PlaySound method, it seemed to resolve the issue. not sure if this is the best way to solve this.


Assets/_PartyParkade/Script/Managers/AudioManager.cs
```diff
@@ -235,7 +235,7 @@ private void PlaySceneSoundConnection()
         {
             StopAmbientNoises();
             SetUpBackgroundMusicConnections();
-            if (m_activeScene.buildIndex != GENERIC_UI_BUILD_INDEX && m_activeScene.buildIndex != -1)
+            if (m_activeScene.buildIndex != GENERIC_UI_BUILD_INDEX)
```
m4rtint, line 238: https://github.com/PerigonGames/Party-Parkade/pull/294/files#r454006391


Not saying this is the best solution, but need to figure out a solution that will cover this as well

PR #312, opened 2020-07-22
BugFix/ItemPickUp

m4rtint:


m4rtint:
[Never got switch + text and drive powerup](https://trello.com/c/PgNtokDu/741-never-got-switch-text-and-drive-powerup)

PR #314, opened 2020-07-23
BugFix/Score Saving

m4rtint:


m4rtint:
[Points selection not active unless Points selection UI is hovered](https://trello.com/c/9GwJqBMU/716-points-selection-not-active-unless-points-selection-ui-is-hovered)

PR #315, opened 2020-07-24
Update player placements in-game

mikeslgoh:
The current implementation works but not sure if it's currently the best way to do it so would like some feedback on this


Assets/_PartyParkade/Prefabs/Player/Player.prefab
```diff
@@ -870,6 +870,11 @@ PrefabInstance:
   m_Modification:
     m_TransformParent: {fileID: 1987924785778101457}
     m_Modifications:
+    - target: {fileID: 740624528853669045, guid: 656c4aa2e60b7489bae3472e003c8ac9,
```
mikeslgoh, line 873: The change for the Player prefab is adding the text mesh object for the placement as a serialized field

mikeslgoh:
[Update player placement based on their in game score (2.0)](https://trello.com/c/TryTAHn2/735-update-player-placement-based-on-their-in-game-score-20)

m4rtint:
What else is there to change?

PR #316, opened 2020-07-25
Feature/scored animations

Gursaanj:
Created 3 new scored animations that will play at random when a player has scored a point

Scored Animations


Gursaanj:
[Create animation for score indicator phrases (3.0)](https://trello.com/c/yw94Hyem/680-create-animation-for-score-indicator-phrases-30)

Assets/_PartyParkade/Script/Game/ScoredPopUpBehaviour.cs
```diff
@@ -15,14 +17,28 @@
...
+    private readonly Vector3 StartingRotationRectPosition = new Vector3(0, 65, 0);
+    private readonly Vector3 StartingBounceBackPosition = new Vector3(0, 115, 0);
+    private readonly Vector3 StartingSlideToRightPosition = new Vector3(-120,65,0);
+    
+    private readonly Vector3 PunchPositionMovement = new Vector3(0, 20, 0);
+    private readonly Vector3 FullCircleRotation = new Vector3(0,0,360);
     
     [SerializeField] private Sprite[] m_wordBank = null;
     [SerializeField] private Image m_scoredImage = null;
+    [SerializeField] private RectTransform m_rectTransform = null;
```
m4rtint, line 36: I already placed a [RequireComponent(typeof(CanvasGroup))]

So this behaviour/script will never work unless you have a CanvasGroup attached to it. 
You won't need to Serialized the field, you just need a GetComponent. 
You can apply the same thing to RectTransform. 

Put in RequireComponent for RectTransform, so you don't need to Serialize it, and just cache a RectTransform OnAwake

m4rtint: And the more you serialize at this point, the higher the chance to get a merge conflict on a prefab.

Gursaanj: hmm I guess that works, but seems like unnecessary getcomponent calls  when its less expensive to serialize, plus I cant imagine anyone else working on the ScoredPopup  prefab at the moment to cause a conflict

But sure I can do that

Gursaanj:
> What did you change for the Player Prefab?

I had to change the width and height of the rect transform of the scored popup in reference to player prefab. So that if the scored popup prefab needs to be used anywhere else (even thought highly unlikely) the standard numbers are still there

Assets/_PartyParkade/Script/Game/ScoredPopUpBehaviour.cs
m4rtint: GetComponent is only called once in the beginning. It shouldn't really effect anything. 

but if u put it in update, that's a problem.

This is more a of design thing in terms of work flow.

PR #319, opened 2020-07-27
BugFix/Prop Reset

m4rtint:


m4rtint:
[Map interactions does not reset on restart if mid animation](https://trello.com/c/8V6IDu2o/766-map-interactions-does-not-reset-on-restart-if-mid-animation)

PR #321, opened 2020-07-30
BugFix/Main Menu BGM

DJSteins:
It fixed


DJSteins:
[Main menu music not there (1.0)](https://trello.com/c/Kd98Mc7R/738-main-menu-music-not-there-10)

Assets/_PartyParkade/Script/Managers/AudioManager.cs
```diff
@@ -280,6 +281,19 @@ private void SetupSceneSoundConnection(Scene scene, LoadSceneMode mode)
             {
                 m_activeScene = scene;
             }
+            if (scene.name == SCENENAME.DoozyMainMenu.ToString())
```
m4rtint, line 284: can you change the conditional to check the index instead? 
so scene.index == 0

DJSteins: Is scene index more likely to remain unchanged or is the scene name more likely to remain unchanged?

m4rtint: scene index. Especially main menu since it'll always be 0 - the scene you first see.

Gursaanj: hmm, if/when we put a splashscreen into the game, wouldnt that be scene index 0, or would that just be the same main menu Scene?? @m4rtint  

If that becomes the new scene index 0, then going by name would be safer

m4rtint: Hmm that's a good point, yeah the splash would be a new scene, since when going from game back to main menu, you wouldn't wanna play the splash (unless we wanna add some logic).

But the name of the Main Menu will be changed too. DoozyMainMenu -> MainMenu. 

I think what we have is fine for now, but then when adding in splash screen, we need to refactor a few things, 
e.g. GenericGameUI needs to be shifted to scene index 2, or do a name compare.

Gursaanj: Sounds good

PR #326, opened 2020-07-31
Feature/powerup manager

Gursaanj:
made the PowerUpObject pooler only pool Boost Powerups when playing in single player mode. Through the help of a PowerUpManager

PowerUpManager


Assets/_PartyParkade/Script/Managers/PowerupManager.cs
```diff
@@ -0,0 +1,21 @@
...
+//-----------------------------------------------------------------------
+
+using _PartyParkade.Script.Data;
+using UnityEngine;
+
+public class PowerUpManager : MonoBehaviour
+{
+    public void Initialize(int numberOfPlayers)
+    {
+        PowerUpsObjectPooler powerUpPooler = FindObjectOfType<PowerUpsObjectPooler>();
```
Gursaanj, line 14: The reason I do a Find here is because the pooler already exists as a prefab in the scene Hierarchy, I know its expensive, so I could serialize the pooler gameobject and do an addcomponent instead, but then the pooler prefab would be a child of the worldmanager prefab, if we are okay with that

Gursaanj:
[Powerup pickup should only give boost (2.0)](https://trello.com/c/QRI9COnc/732-powerup-pickup-should-only-give-boost-20)

Assets/_PartyParkade/Script/Managers/PowerupManager.cs
m4rtint: 1. Honestly, As one Init function in the beginning, I think it's okay. I haven't properly tested out the load repercussion of FindObject, but it should be fine at this point. 

2. why would you need to do an addcomponent? the PowerUpsObjectPooler is already an component of the gameobject. You would just need to create a serialize field, drag drop, and then initialize the pooler below.

Assets/_PartyParkade/Script/Data/ObjectPooler/PowerUpsObjectPooler.cs
```diff
@@ -76,14 +81,32 @@ public static bool IsDebug()
...
+                if (!m_poolDictionary.ContainsKey(tag))
+                {
+                    Debug.LogWarning(message: $"{tag} does not exist");
+                    return null;
+                }
 
-            PowerUp newObj = m_poolDictionary[tag].Dequeue();
-            m_poolDictionary[tag].Enqueue(newObj);
+                newObj = m_poolDictionary[tag].Dequeue();
+                m_poolDictionary[tag].Enqueue(newObj);
```
m4rtint, line 108: ` newObj = m_poolDictionary[tag].Dequeue();
            m_poolDictionary[tag].Enqueue(newObj);	                m_poolDictionary[tag].Enqueue(newObj);`

you could put this outside the if statement
and then do a 
`tag = m_singlePlayerPool.Tag` to store the singleplayerPool tag

Assets/_PartyParkade/Script/Managers/PowerupManager.cs
m4rtint: Either solution is fine, it's up to you.

Gursaanj: For sure makes sense, will leave as is then!!

PR #327, opened 2020-07-31
Bugfix/scored animations fix

Gursaanj:
Changed some of the values for the 3 different scored animations to better align with UI/UX design.

Currently the words "NICE" and  "GREAT" are considerably bigger than the rest, due to some spacing issue, that will be fixed at a later time after getting some refined images from figma

https://perigongames.slack.com/archives/CKJDYR68L/p1595919172028000?thread_ts=1593835383.016000&cid=CKJDYR68L


mikeslgoh:
[Score Indicator effect is there but words are small](https://trello.com/c/QlaN7R8o/765-score-indicator-effect-is-there-but-words-are-small)

PR #329, opened 2020-08-01
Changes made

Gursaanj:
Made changes to the number of Powerups available per PowerUpPool based on the feedback given in the last playtesting session.

Also changed the duration of the OilSpill Powerup based on feedback


Gursaanj:
[Powerup tweaks](https://trello.com/c/AXpCqhoJ/780-powerup-tweaks)

Assets/_PartyParkade/Prefabs/PowerUps/OilSpill.prefab
```diff
@@ -9523,8 +9523,8 @@ MonoBehaviour:
   m_EditorClassIdentifier: 
   m_powerUpProperties: {fileID: 11400000, guid: 325f66602984949fbaff13d3a2cde938,
     type: 2}
-  m_oilSpillTime: 4
-  m_oilPoolDisappearTime: 3
+  m_oilSpillTime: 3
+  m_oilPoolDisappearTime: 2
```
m4rtint, line 9527: Honestly I think the Disappear time to 0.5 is better, seeing how over powered it is.

Assets/_PartyParkade/Scenes/Eight3D.unity
```diff
@@ -98,7 +98,7 @@ LightmapSettings:
     m_TrainingDataDestination: TrainingData
     m_LightProbeSampleCountMultiplier: 4
   m_LightingDataAsset: {fileID: 0}
-  m_UseShadowmask: 0
+  m_UseShadowmask: 1
```
m4rtint, line 101: can u revert this? unless it's intentional?

Assets/_PartyParkade/Prefabs/Managers/PowerUpObjectPooler.prefab
```diff
@@ -47,30 +47,30 @@ MonoBehaviour:
   - m_tag: 0
     m_prefab: {fileID: 2064773118984300792, guid: a9a754b2f715b27499c1c27541a1f662,
       type: 3}
-    m_size: 5
+    m_size: 4
```
m4rtint, line 50: I think making the Size 5 or 6 is better, rather overcompensate.

m4rtint: For all of them.

Assets/_PartyParkade/Scenes/Eight3D.unity
Gursaanj: Yeah I think it is necessary. so long story short the PowerUpManager was originally called PowerupManager (before being renamed) and that was what was being referred to in the scene (not sure why). So I had to remove and readd the same script to the scene, causing this change

Assets/_PartyParkade/Prefabs/Managers/PowerUpObjectPooler.prefab
Gursaanj: hmm but since there cant be more than 4 people per game, adding more would be redundant no??

m4rtint: but if you had 4 bouncy objects, all put out at once, and then all 4 grabbed power up, and it's another 4 bouncy objects. Then problems will happens.

m4rtint: Not saying it'll always happen, but there's always that chance that more than 4 will be needed so better to cover those edge cases.

m4rtint: Just to reiterate, the amount of objects spawning has nothing to do with the Randomization chances on which power up is chosen

Gursaanj: yup thats going to be handled in a different PR , after discussing how to go about it at Today's dev meeting

Assets/_PartyParkade/Prefabs/PowerUps/GravityWell.prefab
```diff
@@ -49,10 +49,9 @@ MonoBehaviour:
   m_EditorClassIdentifier: 
   m_powerUpProperties: {fileID: 11400000, guid: 0ebad3d223156d742ac8e99e8ebc61bd,
     type: 2}
-  m_durationOfPowerup: 7
-  m_heightAboveCarToRayCastBackwards: 3
```
m4rtint, line 53: This is strange that this is removed, can you check if this variable is still set in the prefab?

Gursaanj: Oh yeah that is so weird, cause its on the Deployable class as a private read only thing so im not sure why it would be on the next to all the other serailze fields. Looking at the version control stuff, it doesnt even look like it was ever a serialize field before either

Gursaanj: image

m4rtint: okay that's fine

PR #338, opened 2020-08-06
Working Weighted Randomizer for Powerups

Gursaanj:
Randomization for Powerups (on Pickup) now works on a weighted basis. Where the weight of each powerup can be changed on the respected ScriptableObject. Currently only SwitchPlayers has been changed to a weight of 0.33 where everything is at 1.0


Assets/_PartyParkade/Script/Data/ObjectPooler/PowerUpsObjectPooler.cs
```diff
@@ -104,16 +106,80 @@ public PowerUp SpawnFromPool(POWERUPS tag)
...
+            m_poolDictionary[tag].Enqueue(newObj);
+
+            return newObj;
+        }
+        
+        /// <summary>
+        /// Generates Random POWERUPS using the given weights for each POWERUPS
+        /// </summary>
+        /// <returns></returns>
+        private POWERUPS GenerateWeightedRandomPowerUpTag()
```
m4rtint, line 151: This method currently only works in our specific case of wanting TnD to be lower chance, it won't work if you change any other weights to be something else. 

The weighted stuff i'm referring to is similiar to this algo
https://stackoverflow.com/questions/46563490/c-sharp-weighted-random-numbers

Assets/_PartyParkade/Script/Data/ObjectPooler/PowerUpsObjectPooler.cs
```diff
@@ -104,16 +106,80 @@ public PowerUp SpawnFromPool(POWERUPS tag)
 
             return newObj;
         }
+        
+        /// <summary>
+        /// Spawn Random (weighted) PowerUp object from the pool
+        /// </summary>
+        /// <param name="tag">Enum Key</param>
+        /// <returns>Power Up Object</returns>
+        public PowerUp SpawnRandomPowerUpFromPool()
```
m4rtint, line 115: Honestly, I don't think the ObjectPooler should not be the one randomizing and giving out the Tags. 

ObjectPooler should have one sole job, to pool the object. When you ask for a specific thing, you give a specific thing. That's it. 

You can do the randomization in the new PowerUPmanager. Since the PowerUpManager should already have all the PickUpPower objects?  Which mean u'll need to do a GameObject.FindObjectsWithType(ItemPickUp), where you can pass in the PowerUp Generated.

m4rtint: sort of like the parkingmanager

mikeslgoh:
[Weighted randomization of powerups (1.0)](https://trello.com/c/5xVNeIRm/792-weighted-randomization-of-powerups-10)

PR #342, opened 2020-08-10
Feature/clean up residential

wbeltran777:
- Imported all new models to the project
- Added new buildings to residential
- Added destructable behaviour to many obstacles
- Prefabbed lots of models 
- Minor tweak in Destructibles.cs so that clean up sets objects to their initial rotation instead of identity quaternion
- Added some more parking spots and changed the spawn locations
- Paint brush terrain tool wasn't working with the trees due to how the trees are imported into the project. Dragged them in manually for now to show how they would look, and we can reassess whether we need to use the terrain paint brush tool.


m4rtint:
image
Can you clean up the hierachy a bit? Move the Managers under Manager, and props under Enviornment,

PR #348, opened 2020-08-14
BugFix/DontDestroyOnLoad for AudioManager

DJSteins:
Fixes the bug where AudioManager is null and prevents most sound fx from being played in game. This happens because the AudioManager was getting destroyed when a level was loaded.


mikeslgoh:
[Sound effects don't work if you go through whole UI flow](https://trello.com/c/CgIQEbxU/808-sound-effects-dont-work-if-you-go-through-whole-ui-flow)

PR #349, opened 2020-08-14
bgm/Fade between matches

DJSteins:
Also added obstacle collision sound mixing


m4rtint:
So the Buildings will be obstacles, do you think having the destructible props be another tag for another type sound be useful?

DJSteins:
> So the Buildings will be obstacles, do you think having the destructible props be another tag for another type sound be useful?

I think we should either decide to use one sound for all obstacle collisions, or find different sound effects depending on the type of object being collided with (destructible or not). Obviously the latter will take much more time than the former, and would probably also require refactoring to make it easier to add sound fx to the mixer.

m4rtint:
> > So the Buildings will be obstacles, do you think having the destructible props be another tag for another type sound be useful?
> 
> I think we should either decide to use one sound for all obstacle collisions, or find different sound effects depending on the type of object being collided with (destructible or not). Obviously the latter will take much more time than the former, and would probably also require refactoring to make it easier to add sound fx to the mixer.

Alright we'll stick with the obstacles one for now, since in won't be an easy add on the AudioManager side of things.

mikeslgoh:
[BGM should fade out at the end game (1.0)](https://trello.com/c/fB790rrI/807-bgm-should-fade-out-at-the-end-game-10)

PR #357, opened 2020-08-16
bugfix/extract materials from assets

mikeslgoh:
Most of the buildings on residential were invisible when the Github build finished so here's the fix


ProjectSettings/QualitySettings.asset
```diff
@@ -201,7 +201,7 @@ QualitySettings:
     skinWeights: 4
     textureQuality: 0
     anisotropicTextures: 0
-    antiAliasing: 4
```
Gursaanj, line 204: I keep getting this auto changed too, I think should stay at 4 for AA to work, what do you think @m4rtint ?

m4rtint: yeah should remove this. keep it at 4

mikeslgoh: didnt realise I commit this. I'll switch it back to 4

PR #358, opened 2020-08-16
Powerups/SwitchPlayer

m4rtint:
- There was an issue with the rendering of the car in the UI Main Menu Canvas view. 
- I had to turn off a part of the shader while in UI, which is what the extra code is for.

Basically 
material.SetFloat("Dissolve_Rate", 0) : this is basically how you set parameters on the shader. 

- Components to set: Car, Hat, 4x wheels


Assets/_PartyParkade/Script/Game/Car/CarAppearanceBehaviour.cs
```diff
@@ -29,12 +36,30 @@ public class CarAppearanceBehaviour : MonoBehaviour
...
+                m_materialsToDissolve.Add(skinned.material);
+            }
+
+            if (mesh != null)
+            {
+                m_materialsToDissolve.Add(mesh.material);
+            }
+            var wheelMaterials = wheels.ToList()
+                .ConvertAll(wheel => wheel.gameObject.GetComponentInChildren<MeshRenderer>().material);
+            m_materialsToDissolve.AddRange( wheelMaterials);
```
Gursaanj, line 61: Maybe since System.Linq is an expensive import, using it just here might be unnecessary 

Maybe do 
`
foreach (Transform wheel in wheels)
{
      MeshRenderer wheelMeshRenderer = wheel.gameObject.GetComponentInChildren<MeshRenderer>();
      
      if (wheelMeshRenderer != null && wheelMeshRenderer.material != null)
      {
             m_materialsToDissolve.Add(wheelMeshRenderer.material);
      }
}`

Gursaanj: Up to you though!

Assets/_PartyParkade/Script/Game/Car/CarAppearanceBehaviour.cs
```diff
@@ -44,6 +69,35 @@ public void SetCarHat(PlayerInformation information, Transform carHatTransform)
             carHatTransform.transform.localPosition = GetCarModel(information.StatPointer).CarHatPosition;
         }
 
+        /// <summary>
+        /// Reset Appearance
+        /// </summary>
+        public void Reset()
+        {
+            foreach (Material mat in m_materialsToDissolve)
```
Gursaanj, line 77: null check m_materialsToDissolve just to be safe (as well as mat, but that should never be null, so its fine)

Assets/_PartyParkade/Script/Game/Car/CarAppearanceBehaviour.cs
```diff
@@ -44,6 +69,35 @@ public void SetCarHat(PlayerInformation information, Transform carHatTransform)
...
+        /// <summary>
+        /// Dissolve animation of car
+        /// </summary>
+        /// <param name="isDissolvingOut">is it dissolving in</param>
+        /// <param name="duration">aniamtion time</param>
+        /// <returns>Tween</returns>
+        public Tween DissolveCar(bool isDissolvingOut, float duration)
+        {
+            Tween tween = null;
+            foreach (Material mat in m_materialsToDissolve)
```
Gursaanj, line 92: null check here m_materialsToDissolve

Assets/_PartyParkade/Script/UI/MainMenu/PlayerSelection/3DModelView/CarSelectionMeshViewController.cs
```diff
@@ -3,6 +3,8 @@
 //     Copyright (c) PERIGON GAMES. 2019 All Rights Reserved
 // </copyright>
 //-----------------------------------------------------------------------
+
+using System.Linq;
```
Gursaanj, line 7: what is this being used for?

Gursaanj:
Looks like there are conflicts with the new hat materials

Assets/_PartyParkade/Script/UI/MainMenu/PlayerSelection/3DModelView/CarSelectionMeshViewController.cs
m4rtint: To extract all the materials from the gameobjects

m4rtint:
> Looks like there are conflicts with the new hat materials

It's a material/shader thing

Assets/_PartyParkade/Script/Game/Car/CarAppearanceBehaviour.cs
m4rtint: Most of the times, LINQ will be a bit slower because it introduces overhead. Do not use LINQ if you care much about performance. Use LINQ because you want shorter better readable and maintainable code.

- Basically this is why i used Linq

Assets/_PartyParkade/Script/Game/Car/CarAppearanceBehaviour.cs
m4rtint: m_materialsToDissolve It's never nulled, it always has something. I don't think i need to null check it here.

m4rtint: At the very top I initialized it at compile time. it should be fine

Assets/_PartyParkade/Script/UI/MainMenu/PlayerSelection/3DModelView/CarSelectionMeshViewController.cs
Gursaanj: image
 Doesnt look like its being used,

m4rtint: Remove it.

PR #363, opened 2020-08-18
Added all bouncy objcts

m4rtint:
ScreenCapture


m4rtint:
[Import new bouncy objects (0.5)](https://trello.com/c/QAzamAS5/833-import-new-bouncy-objects-05)

PR #366, opened 2020-08-19
Feature/RaceCarModel

m4rtint:
- Imported correct mesh rotation/position of cars


m4rtint:
[Import new racecar model (0.5)](https://trello.com/c/3vnk0MYK/834-import-new-racecar-model-05)

PR #369, opened 2020-08-19
Scored popup animations with correct sizes

Gursaanj:
Replaced old scored popup sprites with ones with the appropriate sizing
ScreenCapture


Gursaanj:
[Score Indicator effect is there but words are small (0.5)](https://trello.com/c/QlaN7R8o/765-score-indicator-effect-is-there-but-words-are-small-05)

PR #372, opened 2020-08-23
(Enhancement + Bugfix)/Parking Detection

DJSteins:
1. Parking detection wasn't using the proper corners of the car's bounding box, so I switched it to use wheel coordinates instead
2. Reduce the size of the bounding box for parking spots
3. Removed the ParkingDetection class and its unit tests because they aren't really needed for just checking if the wheels are in the parking spot bounds.
4. Added a parking constraint on car speed so that you can't really be parked by just driving through the spot. This could be changed by also just increasing the parking timer.


mikeslgoh:
[Look into better ways to implement the parking detection (2.0)](https://trello.com/c/R48GYPoq/815-look-into-better-ways-to-implement-the-parking-detection-20)

Assets/_PartyParkade/Script/Game/ParkingDetectionBehaviour.cs
```diff
@@ -18,17 +20,21 @@ namespace _PartyParkade.Script.Game
     [RequireComponent(typeof(ParkingTimer))]
     public class ParkingDetectionBehaviour : MonoBehaviour
     {
+        [SerializeField]
+        [Range(0, 100)]
+        private float m_requiredAreaIntersectionPerctentage = 95.0f;
```
m4rtint, line 25: This is not used anywhere?

Assets/_PartyParkade/Script/Game/ParkingDetectionBehaviour.cs
```diff
@@ -51,23 +57,46 @@ private void OnTriggerStay(Collider collision)
...
+                {
+                    return false;
+                }
+            }
+            return true;
+        }
+
+        private void OnDrawGizmos()
+        {
+            if (m_car == null)
```
m4rtint, line 89: You should probably have a flag to turn on or off the Gizmos

DJSteins: Do we though since gizmos are just turned on/off in the editor?

Assets/_PartyParkade/Script/Game/ParkingDetectionBehaviour.cs
DJSteins: Oops

Assets/_PartyParkade/Script/Game/ParkingDetectionBehaviour.cs
m4rtint: In this scenario we can do it through the editor, but sometimes we'll have separate gizmo for different things.
 This is up to you.

Assets/_PartyParkade/Script/Game/ParkingDetectionBehaviour.cs
```diff
@@ -9,6 +9,8 @@
 using _PartyParkade.Script.Data;
 using Sirenix.OdinInspector;
```
Gursaanj, line 10: This import seems to be unused

Assets/_PartyParkade/Script/Game/ParkingDetectionBehaviour.cs
```diff
@@ -9,6 +9,8 @@
...
 using UnityEngine;
+using Sirenix.Utilities;
+using System.Linq;
```
Gursaanj, line 13: same here, seem to be unused

Assets/_PartyParkade/Script/Game/ParkingDetectionBehaviour.cs
Gursaanj: since OnDrawGizmos can only be used in the editor and not in the builds you can wrap the method with the UNITY_EDITOR conditional . for optimization sake

Gursaanj: `#if UNITY_EDITOR
private void onGizmos
{blah}
#endif`

PR #375, opened 2020-08-25
UI/Power Up Error

m4rtint:


m4rtint:
[Power Up Error (0.5)](https://trello.com/c/xbUAoW3v/853-power-up-error-05)

m4rtint:
[Implement UI for powerups not being allowed to be deployed (1.0)](https://trello.com/c/EJBUFuRM/847-implement-ui-for-powerups-not-being-allowed-to-be-deployed-10)

Assets/_PartyParkade/Script/UI/PlayerInGameUserInterface/PlayerPowerUpView.cs
```diff
@@ -22,9 +27,12 @@ public class PlayerPowerUpView : MonoBehaviour
         
         private readonly Vector2 m_rightMinAnchor = new Vector2(0f, 0.7f);
         private readonly Vector2 m_rightMaxAnchor = new Vector2(0.95f, 1f);
+        private readonly Color m_ErrorColor = new Color(1, 0, 0);
```
Gursaanj, line 30: this should be renamed  to m_errorColor to be consistent

Assets/_PartyParkade/Script/UI/PlayerInGameUserInterface/PlayerPowerUpView.cs
```diff
@@ -22,9 +27,12 @@ public class PlayerPowerUpView : MonoBehaviour
...
         
+        [Title("Components")]
```
Gursaanj, line 32: up to you, but if this is the only reason Sirenix Import is being used for, maybe just use [Header("components")] instead and get rid of the import

Assets/_PartyParkade/Script/UI/PlayerInGameUserInterface/PlayerUserInterfaceView.cs
```diff
@@ -74,7 +74,7 @@ public void SetPowerUpIcon (PowerUpProperties properties)
         /// </summary>
         public void AnimateUnableToActivatePowerUp()
```
Gursaanj, line 75: Looks like this method is only ever called from the player controller. Would it be better to have an event in PlayerController.cs that PlayerPowerUpView.cs subscribes and unsubscribes from?. So that PlayerController can just invoke the event instead of call this method which in turn, just calls another method from another class

Gursaanj: Up to you though!

m4rtint: it already have direct access, might as well.

Assets/_PartyParkade/Script/UI/PlayerInGameUserInterface/PlayerPowerUpView.cs
m4rtint: Updated

PR #379, opened 2020-08-27
BugFix/GravityWell

m4rtint:
- Freeze rotation of Gravity well.

This is just one of the solutions, see if this works first. So it can't flip


m4rtint:
[Gravity well gets deployed like a piece of paper](https://trello.com/c/19fhhlRv/821-gravity-well-gets-deployed-like-a-piece-of-paper)

DJSteins:
Did you test launching the gravity well on a ramp? Does it conform to the angle of the ramp?

m4rtint:
> Did you test launching the gravity well on a ramp? Does it conform to the angle of the ramp?

Oh i'll need to test that out.

PR #381, opened 2020-08-28
Bugfix/Options Panel Saving

DJSteins:
1. Had to disable UI buttons on reset and save buttons for the submit/onclick calls to work properly
2. Had to put in explicit navigation for save button because if you try to scroll down from those buttons with the joystick it navigates to something off of the options panel.


DJSteins:
[Cant apply lower volume change](https://trello.com/c/QNXPnQWw/850-cant-apply-lower-volume-change)

mikeslgoh:
What happens if you remove the UI button component? Since we are disabling it, might as well just remove it?

m4rtint:
@DJSteins 

No need to disable, 
Make the  InputMode = None
That should work


Screen Shot 2020-08-27 at 11 16 27 PM

PR #382, opened 2020-08-28
Bugfix/Persisting SFX

DJSteins:
In relation to Kevin's bug about the fireplace sound, I have added a "mute all SFX" and "unmute all SFX" call on certain state changes. Hopefully this will resolve the issue of sfx persisting outside of gameplay when we don't want them anymore. This will require some QA to make sure it isn't accidentally shutting off SFX when it shouldn't be.


DJSteins:
[Fireplace bug](https://trello.com/c/x2O5u8xI/852-fireplace-bug)

PR #386, opened 2020-08-29
BugFix/PowerUpError

m4rtint:
- Can only shake if NOT Tweening


m4rtint:
[Implement UI for powerups not being allowed to be deployed (1.0)](https://trello.com/c/EJBUFuRM/847-implement-ui-for-powerups-not-being-allowed-to-be-deployed-10)

PR #388, opened 2020-08-29
Enhancement/car gravity

DJSteins:
Make car go
image


mikeslgoh:
[Car jitters when driving off a raised plane (1.0)](https://trello.com/c/omf3JGns/857-car-jitters-when-driving-off-a-raised-plane-10)

PR #389, opened 2020-08-29
Bugfix/Skidmarks and Smoke

DJSteins:
Just needed to fix up the OnGround method and change the angle requirement


mikeslgoh:
[Car particle effects don’t appear when making turns](https://trello.com/c/mFvJGoh8/813-car-particle-effects-dont-appear-when-making-turns)

PR #391, opened 2020-08-30
Bug fix/save record high time

Gursaanj:
Made some fixes to make sure that when highscores are set they are displayed and saved correctly!


Gursaanj:
[Single player mode does not save previous best score (1.0)](https://trello.com/c/Q7zvNd8G/825-single-player-mode-does-not-save-previous-best-score-10)

Gursaanj:
[Time Trial Mode : Restarting the map will stop the timer from counting up](https://trello.com/c/7EAZZxNc/841-time-trial-mode-restarting-the-map-will-stop-the-timer-from-counting-up)

Assets/_PartyParkade/Script/Managers/TimeManager.cs
```diff
@@ -63,10 +60,13 @@ public void Reset()
...
-
-    private void OnEndGame()
+    
+    /// <summary>
+    /// Returns game time taken to complete the level clamping to the Max time possible
+    /// </summary>
+    public float GetCurrentTimeClamped()
     {
-        m_singlePlayerHighScore.SaveBestTimeIfNeeded();
+        return m_currentGameTime >= MaxTimePossiblePerMatch ? MaxTimePossiblePerMatch : m_currentGameTime;
```
m4rtint, line 69: How come you don't clamp it, if the method is called Clamp?
like Mathf.Clamp(m_currentGameTime, 0, MaxTimePossiblePerMatch)

Assets/_PartyParkade/Script/Managers/TimeManager.cs
```diff
@@ -63,10 +60,13 @@ public void Reset()
...
     }
-
-    private void OnEndGame()
+    
+    /// <summary>
+    /// Returns game time taken to complete the level clamping to the Max time possible
+    /// </summary>
+    public float GetCurrentTimeClamped()
     {
-        m_singlePlayerHighScore.SaveBestTimeIfNeeded();
```
m4rtint, line 69: How come you moved the saving part from here, to SinglePlayerScorePanel Initialize()?

Gursaanj: Because before, botht he time manager and the SinglePlayerScorePanel were Initializing SinglePlayerHighScore, in which they were setting the player pref key differently. so when it came time for calling GetBestTime() , there was no playerpref with given key and you would always get back the current time (thus the bug).

PR #392, opened 2020-08-30
BugFix/ControllerJoinUI

m4rtint:
Set the auto sizing , lowest Fontsize to be 5 instead of 25


m4rtint:
[The joined text overflows from the box for PS4](https://trello.com/c/uLKFStTG/809-the-joined-text-overflows-from-the-box-for-ps4)

PR #394, opened 2020-08-31
BugFix/OilSpill

m4rtint:
- Oil Spill went through the surface of winterwonderland
- Changed the Collision inside particle effects to also hit the Layer "Obstacle"


m4rtint:
[Deploying oil drop powerup on Winterland hills and patches will drop below *ground* level](https://trello.com/c/WCaCT48U/861-deploying-oil-drop-powerup-on-winterland-hills-and-patches-will-drop-below-ground-level)

PR #396, opened 2020-08-31
BugFix/StartupErrors

m4rtint:
Set State to execute first


m4rtint:
[Error Log at game startups](https://trello.com/c/A2ibIItH/863-error-log-at-game-startups)

PR #398, opened 2020-09-03
Enhancement/killing plane

Gursaanj:
Created an invisible box collider around the map that will inactivate any destructible item 

that passes through it or respawns the player to the appropriate spawn point

ScreenCapture


Gursaanj:
[Out of bounds respawn (0.5)](https://trello.com/c/1tO1nuOI/805-out-of-bounds-respawn-05)

Assets/_PartyParkade/Script/Game/KillingFloor.cs
```diff
@@ -0,0 +1,58 @@
+﻿//-----------------------------------------------------------------------
+// <copyright file="KillingFloor.cs" company="PERIGON GAMES">
+//     Copyright (c) PERIGON GAMES. 2020 All Rights Reserved
+// </copyright>
+//-----------------------------------------------------------------------
+
+using _PartyParkade.Script.Data;
+using UnityEngine;
+
+namespace _PartyParkade.Script.Game
```
m4rtint, line 10: Can you name this PerigonGames

Assets/_PartyParkade/Script/Game/KillingFloor.cs
```diff
@@ -0,0 +1,58 @@
...
+            {
+                m_collidedObject.SetActive(false);
+            }
+        }
+        
+        private bool IsPlayerModelLayer()
+        {
+            if (m_collidedObject != null)
+            {
+                return m_collidedObject.layer == LayerMask.NameToLayer(Layers.PlayerModel);
```
m4rtint, line 52: Why use a layer? why not tag it?

m4rtint: Normally we leave Layers to things that is related to physics, like raycasting. 
Not a big deal, just a general rule of thumb. 

> There are 32 layers, but a (more or less) infinite number of tags, since they are strings, whereas layers are bits

Gursaanj: Good point, the reason I did this is because when the car passes through 2 colliders trigger the OnTriggerExit event, one being the CarRigidBody and the other being PhysicsMaterial. To make sure I dont deactivate the PhysicsMaterial, i make this check (since it has a special layer but no special tag).

Gursaanj: I guess Ill just make a new tag and give it to the PhysicsMaterial Child and check that instead

Assets/_PartyParkade/Script/Game/KillingFloor.cs
```diff
@@ -0,0 +1,49 @@
...
+                if (controller != null)
+                {
+                    controller.RespawnPlayer();
+                }
+                else
+                {
+                    DeactivateObject();
+                }
+            }
+            else if (!other.CompareTag(Tags.PlayerMaterial)) // Ensures PhysicsMaterial of Car does not deactivate
```
m4rtint, line 35: Would it be an issue to tag the PhysicsMaterial as a PlayerCar as well?

m4rtint: I didn't double check, i'm sure it would be an issue, but just incase there's the possibility.

Gursaanj: I wanted to do that but CarRigidBody, the child with the other collider (that we want) has that tag. And I dont want to change that, since Im not sure what is using that

m4rtint: You should call this PlayerCarCollidingBox or something more specific. Since it could be mixed up with the shader/material context.

m4rtint: And I just checked your branch, it seems like the physics material still has the layer and not tha tag.

m4rtint: You should also remember to delete the layer you created

Gursaanj: The layer was there before, I didnt add/change anything there

Gursaanj: changed tag name

m4rtint (approved):

ProjectSettings/TagManager.asset
```diff
@@ -11,6 +11,7 @@ TagManager:
   - WaitZone
```
m4rtint, line 11: from what I can see
PlayerModel
RoadConnection
  - WaitZone
  - Gib 

All can be removed. can u remove it in this PR?

m4rtint (changes requested):
Change tag manager

m4rtint (approved):

m4rtint:
neverimind , we can clean eeverything up in one go

mikeslgoh (approved):

PR #402, opened 2020-09-03
Refactor/NameSpace

m4rtint:
- Renamed and refactored all the namespaces to use PerigonGames
- removed as much redundant using as possible
- The big change you see is because of Ryder indentations. 

This is a big change, I ran all the tests and everything seems to run fine. 

No rush to merge this in. But it's a very nice bonus.


Gursaanj:
Should the namespace of everything be PerigonGames? I was under the impression that anything that is easily identifiable as game agnostic (i.e a static utility class or editor tools that can be used in any project) should be classified with PerigonGames and anything game specific (i.e parking or driving in this case) should be PartyParkade NameSpace??

Just a question? I dont mind either way

m4rtint:
> Should the namespace of everything be PerigonGames? I was under the impression that anything that is easily identifiable as game agnostic (i.e a static utility class or editor tools that can be used in any project) should be classified with PerigonGames and anything game specific (i.e parking or driving in this case) should be PartyParkade NameSpace??
> 
> Just a question? I dont mind either way

This is a good point, but I would argue out of consistency it should stay the same. And I don't want to sit here all day wondering down the line, which file should be party parkade, and which should be perigon games for new files/old files.

m4rtint:
Starting from this project, naming everything Perigon Games is fine, but then moving on to another project, we can start naming it namespace {ProjectName} and keeping the tools agnostic.

Gursaanj (approved):

mikeslgoh (approved):

PR #403, opened 2020-09-04
Bug fix/text and drive working emojis

Gursaanj:
Some typo fixes/amendments and deleting the TextAndDrivePreviewScene!


Gursaanj:
[Go through generated conversations for typo pass and add new emojis](https://trello.com/c/prawTdr2/745-go-through-generated-conversations-for-typo-pass-and-add-new-emojis)

mikeslgoh:
Are there emojis added to the conversations?

m4rtint (approved):

mikeslgoh:
> > Are there emojis added to the conversations?
> > Because I remember that Jocelyn mentioned there should only be one spritesheet in the project and I'm not sure if I accidentally override her change when I was cleaning her branch
> 
> So the emojis work, but there are multiple spritesheets. I can delete the unused ones here, or we can save that for when we clean up all redundant assets

So the one that is supposed to be used I believe is spritesheet2? That should include Magan's emojis as well but I think I overrided the change where Jocelyn pointed the script to read from the right file. Might be worth deleting and making the change here

m4rtint:
Just make sure you rename the properly used sprite sheet properly., no spaces, only underscore

mikeslgoh (approved):

PR #406, opened 2020-09-08
Playtesting changes from September 6th

mikeslgoh:
Main change here is that we found that the collider for the StreetParking prefab was off center in some maps. Also made the parking collider wider so that it is slightly easier to park


mikeslgoh:
@wbeltran777 @Gursaanj can I get help double checking a few things?
- Is the parking prefab suppose to have threshold 8 or be the same as the StreetParking prefab?
- If they are the same value, could we revert the change and make the value a const in ParkingBehaviour?

wbeltran777:
> @wbeltran777 @Gursaanj can I get help double checking a few things?
> 
> * Is the parking prefab suppose to have threshold 8 or be the same as the StreetParking prefab?
> * If they are the same value, could we revert the change and make the value a const in ParkingBehaviour?

Yes we set it to 8, and it should be the same on both prefabs. That being said, yes we should probably revert the changes and just keep it as a const

m4rtint:
The scenes are still the old file names, is this not up to date with the current develop

mikeslgoh:
[Look into better ways to implement the parking detection (2.0)](https://trello.com/c/R48GYPoq/815-look-into-better-ways-to-implement-the-parking-detection-20)

wbeltran777 (approved):

m4rtint (approved):

PR #413, opened 2020-09-14
SFX bug fixed

Gursaanj:
Leaving/Restarting a match during a powerup usage will no longer cause the powerup to play forever!


Assets/_PartyParkade/Script/Game/Car/CarAudioSystem.cs
```diff
@@ -53,6 +56,14 @@ private void SetEngineSound(PerigonSoundEffect engine)
             m_wheelVolume = MinWheelVolume;
         }
 
+        private void ResetEngineSound()
```
Gursaanj, line 59: So it turns out that , SoundManager.StopSFX() being invoked on pregame made engine lose its reference to the audiosource  so I had to cache the reference of engine here and reapply it onPregame.invoke

Gursaanj:
[Powerup SFX can be heard on loop forever](https://trello.com/c/PQVMBXQx/890-powerup-sfx-can-be-heard-on-loop-forever)

mikeslgoh (approved):
thanks for cleaning up the audio manager

Assets/_PartyParkade/Script/Game/Car/CarAudioSystem.cs
```diff
@@ -28,6 +28,7 @@ public class CarAudioSystem : MonoBehaviour
         private const float MinWheelVolume = 0.3f;
         private float m_wheelSpeed;
         private float m_wheelVolume;
+        private PerigonSoundEffect engineSoundEffect = PerigonSoundEffect.NoSFX;
```
m4rtint, line 31: Don't need to cache this

m4rtint (approved):

PR #417, opened 2020-09-16
Enabled Autosizing for 3 and 4 player mode

Gursaanj:
Added Autosizing for the Contact names (For Text and Drive) for 3 and 4 player mode. Didnt add it to two player mode because the content looked like it way too much wasted space as a result, so left it be

**Two Player**
TwoPlayer

**Three and four player**
ThreeAndFourPlayer


Gursaanj:
[Resize TextAndDrive Contact Name](https://trello.com/c/Ux6nsXmU/894-resize-textanddrive-contact-name)

Assets/_PartyParkade/Script/PowerUp/DriveText/TextAndDriveTopBarBehaviour.cs
```diff
@@ -144,6 +146,11 @@ private void AssignRandomTopBarElement(TextMeshProUGUI container, string[] possi
     private void Awake()
     {
         OriginalPosition = transform.localPosition += new Vector3(0, AnimationYMovementDownwards, 0);
+
+        if (PerigonGames.GameData.PlayerCount > 2)
+        {
+            m_nameOfContact.enableAutoSizing = true;
```
m4rtint, line 152: How come only autosizing with 2+ players, why not just autoresizing in general.

mikeslgoh (approved):

Assets/_PartyParkade/Script/PowerUp/DriveText/TextAndDriveTopBarBehaviour.cs
```diff
@@ -3,10 +3,12 @@
 //     Copyright (c) PERIGON GAMES. 2019 All Rights Reserved
 // </copyright>
 //-----------------------------------------------------------------------
+
 using DG.Tweening;
 using TMPro;
 using UnityEngine;
 using UnityEngine.UI;
+using Random = UnityEngine.Random;
```
mikeslgoh, line 11: why do you need random for?

Gursaanj: I think this is a rider/git thing. So Random is being used to assign the top bar elements and rider just automatically assigned this

Assets/_PartyParkade/Script/PowerUp/DriveText/TextAndDriveTopBarBehaviour.cs
Gursaanj: because when it auto resizes for two, a lot of wasted space comes out (since the blue bars are much longer when vertical) and it looks pretty bad, in my opinion. So I just left it as is.

Assets/_PartyParkade/Script/PowerUp/DriveText/TextAndDriveTopBarBehaviour.cs
m4rtint: it's not a Git thing. It's a C# thing. it's like type aliasing. 

Anyway I don't see this being used.

m4rtint:
The UX side of things,  the time does not align with the symbols on the right. 
Do you want to make sure they line up? or leave it as it is. 

This is up to you.

Assets/_PartyParkade/Script/PowerUp/DriveText/TextAndDriveTopBarBehaviour.cs
Gursaanj: Nah, I think rider just added it. Will remove it

Gursaanj:
> The UX side of things, the time does not align with the symbols on the right.
> Do you want to make sure they line up? or leave it as it is.
> 
> This is up to you.

good point, will try to align it

m4rtint (approved):

PR #419, opened 2020-09-17
Feature/parking indicator off during countdown

wbeltran777:
- Scaled up indicator arrow, looks good so far (Player prefab changes)
- Updating direction of arrow was inside CalculateScalingSize, so I moved that to a separate function so we could update the arrow without necessarily calling CalculateScalingSize
- Added scaling tween for when indicator is activated (Need to stop the scaling on update for this to happen, which is why I added a bool conditional on the update function)


Assets/_PartyParkade/Script/Game/Car/ParkingIndicatorBehaviour.cs
```diff
@@ -16,7 +16,9 @@ namespace PerigonGames
     public class ParkingIndicatorBehaviour : MonoBehaviour
     {
         private readonly Vector3 m_arrowMovementAmount = new Vector3(0, 3, 0);
-        private const float DurationOfAnimation = 1f;
+        private const float DurationOfArrowBop = 1f;
+        private const float DurationOfIndicatorActivation = 0.7f;
+        private bool m_indicatorIsVisible = false;
```
m4rtint, line 21: in terms of naming conventions , it's usually
isIndicatorVisible

Assets/_PartyParkade/Script/Game/Car/ParkingIndicatorBehaviour.cs
```diff
@@ -46,16 +48,9 @@ private void Awake()
...
-                m_arrowPointing.transform.DOPunchPosition(m_arrowMovementAmount, DurationOfAnimation, 0, 0)
-                    .SetLoops(-1);
-            }
-        }
-
-        private void Start()
-        {
-            SetChildrenActive(true);
+            CleanUpIndicator();
+            StateManager.Instance.OnPlay += InitializeIndicator;
```
m4rtint, line 52: Put the subscription on Start/Awake


since every single time you disable/enable this script, it'll subscribe, but never unsubscribe

Gursaanj (approved):

mikeslgoh:
[Refactor parking indicator shown during countdown](https://trello.com/c/Wx8oBD97/832-refactor-parking-indicator-shown-during-countdown)

Assets/_PartyParkade/Script/Game/Car/ParkingIndicator.cs
```diff
@@ -56,5 +56,10 @@ public float CalculateScalingSize(Transform target, Transform player, Vector3 si
                 return dist / buffer;
             }
         }
+
+        public void UpdateIndicatorDiretion(Transform target, Transform player)
```
m4rtint, line 60: typo

Assets/_PartyParkade/Script/Game/Car/ParkingIndicatorBehaviour.cs
```diff
@@ -105,5 +140,13 @@ private void SetChildrenActive(bool value)
                 child.gameObject.SetActive(value);
             }
         }
+
+        /// <summary>
+        /// Enables indicator scaling according to distance once indicator becomes visible to the user
+        /// </summary>
+        private void IndicatorIsVisible()
```
m4rtint, line 147: Rename this as 
"SetIndicatorAsVisible"

wbeltran777: Good tip!
I'm realizing both names aren't super accurate since this doesn't set the visibility rather it allows the indicator to scale according to the distance to the parking spot. 
So I might use "ActivateDistanceBasedScaling" ?

m4rtint: no, if you used that name, then devs would expect to have "Distance" and "scaling"

but this is just setting a bool to true

m4rtint (approved):

PR #423, opened 2020-09-20
Feature/winter land update

wbeltran777:
- Added a bunch of trees! 
- Repositioned Igloo
- Added christmas market portion
- Added a trainset
- Added Snow park area with benches, gingerbread men and camp fire.
- Ready for hand off for texture painting and for other destructibles to be added if we need them 
WWUpdate


m4rtint:
Screen Shot 2020-09-19 at 11 02 22 PM

I know you used the terrain to put in trees which is great, but there are a few trees you put as gameobject which i'm assuming since you want them to be a different size, but there are a few trees in there which seems to be the same as any other tree on the outskirts. Are you sure you need them as separate gameobjets?

wbeltran777:
> Screen Shot 2020-09-19 at 11 02 22 PM
> 
> I know you used the terrain to put in trees which is great, but there are a few trees you put as gameobject which i'm assuming since you want them to be a different size, but there are a few trees in there which seems to be the same as any other tree on the outskirts. Are you sure you need them as separate gameobjets?

You're right I don't specifically need them as separate objects, will update soon

m4rtint:
Screen Shot 2020-09-19 at 11 07 04 PM

So I only checked the train, but you might want to make a local build for yourself and see if there are any missing materials, since lambert is not normally the name of the material, just unsure a build would be the same

m4rtint:
Screen Shot 2020-09-19 at 11 09 13 PM
1. Since it's a prefab that you created, you can override these properties and components so that if I, Michael or whomever wants to add more trains, the prefab would stay consistent when we choose from the prefab list. 

2. This is more so for next time, but since you already changed the scaling of the transform that's fine. But you can also change the scale factor here so it's more accurate to our map scalings, and then the Transform the scale for fine tuning to other dev's liking, so that's it's not off by a scale of x0.001, or x1000
Screen Shot 2020-09-19 at 11 10 00 PM

wbeltran777:
Addressed most of the issues on the latest commit, will make a build tomorrow morning to verify that the materials are all good

wbeltran777:
Yep great call on the materials, they're gone in the build. 
Latest commit addressed stuff in the comments, I have to step away from my computer for now unfortunately, but I'll fix the material stuff either later tonight or tomorrow morning !

m4rtint (approved):

mikeslgoh (approved):

mikeslgoh:
[Add more details to maps - Winterland (2.0)](https://trello.com/c/WrKvVy4d/846-add-more-details-to-maps-winterland-20)

PR #427, opened 2020-09-23
BugFix/IconDisappear

m4rtint:
- Use scale instead of color changing.
- Nice Animation to make it appear


m4rtint:
[Item Icon turns white when player spams deployable item as they're leaving a restricted deployable area](https://trello.com/c/GQ8G2jpd/913-item-icon-turns-white-when-player-spams-deployable-item-as-theyre-leaving-a-restricted-deployable-area)

mikeslgoh (approved):

wbeltran777 (approved):

PR #439, opened 2020-09-27
Playtesting/09 27

wbeltran777:
- Truck does not go VROOM VROOM now it just goes vroom vroom
- Also camera rotation for three player is slower


m4rtint (approved):

wbeltran777:
[Only truck’s car engine can only be heard at proper volume (1.0)](https://trello.com/c/mTxacBOv/867-only-trucks-car-engine-can-only-be-heard-at-proper-volume-10)

Gursaanj (approved):

PR #441, opened 2020-09-28
Refactor/LoopMapRotation

m4rtint:


m4rtint:
[Camera needs to rotate on loop on every map when playing with 3 players](https://trello.com/c/HVjaLsih/924-camera-needs-to-rotate-on-loop-on-every-map-when-playing-with-3-players)

mikeslgoh (approved):

Gursaanj (approved):

PR #447, opened 2020-09-30
Bugfix/car trail visible on restart

wbeltran777:
- Trail parent is turned on onPreGame and turned back on onPlay

Tried it by only clearing the trails on OnPreGame by using trailRenderer.Clear(), but that still results in some tiny artifacts


Assets/_PartyParkade/Script/Game/Car/CarControllerBehaviour.cs
```diff
@@ -46,6 +46,8 @@ public class CarControllerBehaviour : MonoBehaviour
         private Transform m_body = null;
         [SerializeField] 
         private Transform m_hatPosition = null;
+        [SerializeField]
+        private GameObject m_trailRendererParent = null;
```
m4rtint, line 50: how come you're setting it here? 
Screen Shot 2020-09-30 at 6 56 13 PM

You can do all this in here instead (the smoke particles component

wbeltran777: Yeah I totally overlooked that script
Updated with the changes on this file instead

m4rtint (approved):

Gursaanj:
[Restarting match would create car trails](https://trello.com/c/XJl6VxOe/923-restarting-match-would-create-car-trails)

Gursaanj (approved):

PR #448, opened 2020-10-01
Enhancement/New Sound Effects

m4rtint:


Assets/_PartyParkade/Script/Game/Destructibles/Destructible.cs
```diff
@@ -23,14 +23,21 @@ public abstract class Destructible: MonoBehaviour
...
+        [PropertyOrder(100)]
+        [Title("Audio")] 
+        [SerializeField] private PerigonSoundEffect m_collisionSound = new PerigonSoundEffect();
+        [SerializeField] private float m_activatedSFXVelocity = 5f;
+    
         private Rigidbody m_rigidbody = null;
         private Vector3 m_originalMeshLocalPosition = Vector3.zero;
         private Quaternion m_originalMeshRotation = Quaternion.identity;
 
+        public AudioManager AudioManagerInstance => AudioManager.Instance;
```
Gursaanj, line 35: Does this have to be public?? or will protected be okay?

m4rtint: protected is fine

Gursaanj (approved):
Looks good, I will either get the new SFX in this PR or in another one

m4rtint:
> Looks good, I will either get the new SFX in this PR or in another one

Yeah do it in next PR

mikeslgoh (approved):
I think for playtesting we should play around with the volume because I still don't like the sound of the wall collision

mikeslgoh:
[SFX for destructible objects (2.0)](https://trello.com/c/kl3UKs5n/753-sfx-for-destructible-objects-20)

